PR #65, opened 2019-11-16
Fix machine translation

maboroshin:
I corrected it to natural Japanese.

It became a Japanese version. Thank you!


Armin2208:
Funny how machine translation failed with one word lol. thanks for the correction.

Issue #67, opened 2019-11-27
location service offset


Spiritreader:
Hi,

it would be great to be able to set a custom offset that shifts the sunset and sunrise times dependent on the automated location service.

With multiple night light and dark theme applications I found that for my preference, it often switches too early. 
Preferrably, I would like to switch themes a few minutes / hours after sunrise or sunset. However I also don't want to miss out of having the sunrise and sunset times calculated automatically.

I'd like to propose changes which I have already worked on, as I think this contribution increases QoL a lot, especially for users that are already so picky that they use an auto theme switcher ;)

## The way this is implemented is as follows

Presently it looks like this:
![grafik](https://user-images.githubusercontent.com/18438899/69711208-f1a4c100-1100-11ea-8586-7c02607ebe76.png)

As soon as the location box is ticked, a new set of controls will appear that will either allow a negative or positive offset.

I've tried to adhere to the UI style you used, as I didn't want to disrupt it.

As for how the offset application works: Two offset integers are created in the Settings database, which will be applied each time the location service polls for a new time. Pressing the `set` button in the GUI will also trigger this. 

It would be great if we can start a dialogue on whether it would be possible to merge this.

A pull request will follow shortly which I will reference to this issue.

Thank you!

Armin2208:
❤️🔥

Spiritreader:
Merged. Thanks Armin!

PR #68, opened 2019-11-27
add offset capability for auto location svc

Spiritreader:
#67 pull request that proposes changes for adding offset capabilities for the location service.


Armin2208:
thank you! nice addition.

PR #110, opened 2020-06-23
Resources.ja.resx BETA8 Task function

maboroshin:
Nice evolution.😊


Armin2208:
thanks buddy, still one week of work left until the first release candidate! now they are more strings waiting to be translated :D

Issue #129, opened 2020-07-07
Feature suggestion - TeamViewer dark mode


emberaholdon:
TeamViewer also has a dark mode but unfortunately it does not respect system settings, so Auto Dark Mode could flip its registry key:

HKEY_CURRENT_USER\SOFTWARE\TeamViewer
ColorScheme (DWORD) = 1 (light theme) / 2 (dark theme)

I suspect there are several other apps that are using their own registry keys for their respective color schemes, so the "Apps" section of this tool could be extended greatly. :)

EndPositive:
I would like to help with this. How do we want the apps page to be shaped? I can imagine there will be more apps like this and with the current setup that might become a long page.

emberaholdon:
What first comes to mind is the Nvidia control panel, where there is a dropdown list of the supported apps and then below it are the options for the selected app.

I think something like this would fit into the current UI nicely without needing much layout change:

![app_selector](https://user-images.githubusercontent.com/67831954/87185431-0cbbee00-c2ea-11ea-8758-8976b9be32e3.png)

Armin2208:
I've planned something that allows users to add their own registry keys for apps that support theme switching. We need a more flexible UI that incorporates this and I'm already working on conceptualizing it. 
Still, thank you for your contribution. I can potentially use parts of it to make the overall experience better

emberaholdon:
> I've planned something that allows users to add their own registry keys for apps that support theme switching.

That's a very neat idea.

Armin2208:
For this you can now run your own scripts in Auto Dark Mode Beta.

Check out our wiki page: https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/wiki/How-to-add-custom-scripts

Thanks for your issue and being patient.

Spiritreader:
Custom scripts have been in stable for a while now.
Support for individual programs will probably not make it into ADM and we will refer to our custom scripts instead, unless it's a more complicated endeavor that requires maintenance from our part (if we have the resources).

PR #132, opened 2020-07-10
Update Apps Page interface to allow for more custom apps

EndPositive:
As first seen in #129, a new UI was proposed to allow more custom apps to be added to this page. Please do test and let me know if everything works as you would expect. New translations will need to be done.

### Preview
overview
extra_options


Spiritreader:
This proposal looks pretty good!

You can already add FluentWPF10 to the project.
It's in the directory "nuget" in the repository root.

In Visual Studio by right click on the `AutoDarkModeApp` project -> manage nuget packages. Click the gear icon and add the directory as a package source. Click on "update" and save.
After Visual Studio does it's weird package syncing thing it should work.

image

EndPositive:
Sounds good, I updated the screenshots

Armin2208:
I've planned something that allows users to add their own registry keys for apps that support theme switching. We need a more flexible UI that incorporates this and I'm already working on conceptualizing it. 
Still, thank you for your contribution. I can potentially use parts of it to make the overall experience better

PR #174, opened 2020-10-25
Rename Resources.zhtw.resx to Resources.zh-TW.resx

thomassth:
Accidentally deleted #164.

Same issue.


Armin2208:
although we didn't merged this change, Wie acknowledged it and we got it to work. thanks (and sorry) 🙌🏻

PR #198, opened 2020-12-27
zh-cn: translate offset as "adjustment"

Artoria2e5:
the original translation glosses to "floating (offset)", which doesn't really mean anything. the translator was probably aware of that, hence the parentheses and the english.

translating it literally [偏移(量)] does not seem to help much either, since it sounds like talking about a deviation. programmers might end up figuring that out, but eh...


Armin2208:
thanks for this small change

PR #201, opened 2021-01-09
Fix Chinese translations of Office theme and donation

WordlessEcho:
- alternative should translate to antonym of tradition
- "您好" will be more appropriate than "你好"
- remove the space between the "！" and "感" of "你好！ 感谢"


Armin2208:
great, thanks for supporting this project!

PR #241, opened 2021-07-25
Update Traditional Chinese translation

SiderealArt:


Armin2208:
damn, big update. Thanks for the new PR!

Issue #207, opened 2021-02-04
[Feature suggestion] Allow any offset number


avi12:
I suggested in #139 to take into account weather data
Today turned out to be a [rainy day](https://weather.com/weather/tenday/l/73ac32c92ac8a001c2456904df0ddd2a26e304ad9832ace7811984fea5d62991) all of a sudden, and it turns out that I need a large 3-digit offset, whereas the application will only accept a 2-digit offset

Spiritreader:
already tracked in #180

Issue #180, opened 2020-11-14
Offset digits limit


mburkasov:
When customizing offset for location-based switching cannot input more than two digits, with a maximum of roughly 1.5 hours.

Armin2208:
Can you explain me your usecase why you need more?

mburkasov:
Basicallyt I can't input a number more than 99
![image](https://user-images.githubusercontent.com/20526265/99156166-d6dfa980-26cf-11eb-94cc-2a27dc4c9c76.png)

mburkasov:
> Can you explain me your usecase why you need more?

Well, I wanted an offset the sunset by 3 hours, because the sun sets really early, and I can't do that.

CXwudi:
Same here, I want the dark mode come 4 hours earlier but with this digit limit I can't it

jwfxpr:
Having had a look at this area of the code recently for a related change, and checking again now, I can't see a particular technical reason why the input fields for the offsets must be limited to 2 characters. That said, of course, increasing the fields from 2 to 3 characters would increase the range for the offsets from 1.65 hours each way to 16.65 hours each way, which would allow a silly user to set completely impractical offset values, which I think is the reasoning behind the 2 character limit. 

I think perhaps there could be a middle ground, with an upper limit of, say, 6 or 8 hours on the offset? (For users at extreme latitudes, 6 or 8 hour offsets don't seem that unreasonable to me.)

@Armin2208 Any objection to a change which increases the field length to 3 characters, and adds a validation step on focus loss that would coerce the value to a maximum limit? I can make those changes in the next few days, probably, and open a PR for more specific discussion about implementation or UI changes required for it.

Armin2208:
after we geared so much feedback about this, we want to enable this in future. just didn't had time to implement this. I don't know if there is any conflict with the check numbers for validation code.
feel free to add this change, if you want to.

jwfxpr:
I have an idea of how this could be sensibly done without adding any new regionalised strings. I'll have some time in the next week to play with it a bit 👍

jwfxpr:
Additional discussion of this feature and its implementation can be found in the #243 PR conversation thread.

Spiritreader:
Feature implemented, for more details see f5bffd0 or #243

PR #243, opened 2021-07-27
Increase time offset minute fields to 3 characters

jwfxpr:
Addresses #180 and #207.

This is the simplest viable solution I can see to increase the sunrise/set minutes offset to allow more than 99 minutes. My solution here is:

* Increase the `MaxLength` of the offset fields from 2 to 3 characters, allowing +/- 999 minutes (+/- 16h 39m)
* Since this allows a user to move dark and light times to silly values, if the result means the Dark mode time is earlier than the Light mode time, a small warning glyph is displayed next to the light and dark times in the config screen.

This solution doesn't attempt to coerce or limit the times, and if the user does want those settings, then the app will go ahead and set the dark and light times as configured. This solution as it stands at the moment also doesn't offer any text explanation for the warning or app behaviour, as I wasn't certain you would want to add to the resources for regionalisation and translation.

## Screenshots

Reasonable 3-digit offset:

image

Offset resulting in switched light and dark times:

image

Offset resulting in _really_ switched light and dark times:

image


Armin2208:
Thanks!
I hate this warning symbol. A string would be better.
I think SunsetWithOffset can't be bigger than SunriseWithOffset anyways? Didn't find it in the code, thought it would be the case. I need to work on that.

jwfxpr:
> I hate this warning symbol. A string would be better.

No problem, we can add a warning string to the `Info` feedback section, something like:

    Your current configuration sets Dark mode earlier than Light mode; consider adjusting your configuration.
    Your changes have been saved!

> I think SunsetWithOffset can't be bigger than SunriseWithOffset anyways? Didn't find it in the code, thought it would be the case. I need to work on that.

From what I can see, in 24h time mode it's possible to set a custom Dark time which is earlier than the custom Light time without issue, so this can occur without an offset at all. Is that something you want to change? In that case I'll abandon this PR until that's done. 😊

Spiritreader:
Hey, thanks for making an effort on improving this.
The entire offset section for the interface needs a rework for version X beta 5.0 I believe.
I actually PR'ed the offset feature quite a while ago as a means to get familiar with the project. So it can definitely use some work, especially now!

I believe though, that it is more confusing if you were able to switch day and night cycles.

Therefore a a limit would make sense to prevent switched light and dark times.
However, up to that point, I believe the offset should be freely configurable up to a set limit (1 minute)
So, if users desire, they can adjust their offset such that the day/night cycle is only 1 minute.

jwfxpr:
I can see that this is now part of the [X Beta 5](https://github.com/orgs/AutoDarkMode/projects/1#card-67956034) milestone, so if it's back to being an active priority I'll cheerfully get back to it! 😊

> Therefore a a limit would make sense to prevent switched light and dark times.
> However, up to that point, I believe the offset should be freely configurable up to a set limit (1 minute)
> So, if users desire, they can adjust their offset such that the day/night cycle is only 1 minute.

@Spiritreader, this is doable (of course) without too much trouble, however, it needs a little additional planning. For half the year (from summer solstice to winter solstice) each day will become progressively shorter; so, a user who sets a 1 minute gap on any of these days will possibly find that _the very next day_ their day/night cycle will be switched. It means that:

1. We will need to validate and adjust the offset times at least once each day to continue ensuring that the day/night cycle doesn't get switched around;
2. We will need to decide what method is used to adjust the offset times automatically in these cases. E.g., do we:

- Decide arbitrarily that one of the offsets (light offset or dark offset) is canonical, and the other will always be adjusted;
- Decide that the shorter (for example) of the two offsets is canonical (with an arbitrary tiebreaker) and the other will be adjusted;
- Decide that both offsets will be adjusted by half of the overlap amount;
- Something else I haven't thought of...

Spiritreader:
Hmm yes, I didn't consider that there are edge cases as a result of this.
Updating the offsets periodically in the backend based on suntimes is architectually more difficult than one might assume, and something that is avoided if perferred.

As a design principle, to avoid config synchronization issues between the frontend and backend, the config file is only meant to be written to by the frontend, such that the backend consumes it (mostly) read-only. There's some exceptions based in case a feature needs to be disabled due to an error, but the implications of this when configuring using the frontend are minimal.

But however after giving this issue some more brain time, I have to retract my statement here:
> I believe though, that it is more confusing if you were able to switch day and night cycles.

After thinking about a few scenarios on how to make this work, I am starting to believe that it would actually be more confusing if the software **did affect** the offsets and adjusted them.
Time related stuff is complex enough as is.

> - Decide arbitrarily that one of the offsets (light offset or dark offset) is canonical, and the other will always be adjusted;
> - Decide that the shorter (for example) of the two offsets is canonical (with an arbitrary tiebreaker) and the other will be adjusted;
> - Decide that both offsets will be adjusted by half of the overlap amount;
> - Something else I haven't thought of...

I have attempted to come up with an example scenario for each of your mentioned cases, and it just ends up becoming more complex to explain to the end user.

Thus I believe now that **permitting** switched day/night cycles is fine.

Spiritreader:
Update:

I have talked with @Armin2208 , and for Beta-5 we will probably switch the UI framework.
As such, this draft can only advance once the UI has been restructured.
However, Armin has set a pretty tight release schedule of October 5th, as such this should advance relatively quickly, at least for our pace standards 😅

Spiritreader:
The textboxes have been replaced with ModernWPF numberboxes, which don't have a limit anymore.
This can be considered implemented in f5bffd0, thanks for your efforts nonetheless!

Armin2208:
We removed any time checks, so you are now completely free in your choices.

Issue #256, opened 2021-09-09
Add blue light detection for theme switch


thebigsmileXD:
http://stackoverflow.com/questions/43340619/ddg#43953978

It would be cool to have a mode to detect if the blue light filter is running, as alternative to the time based / location based mode

Armin2208:
oh wow! thanks.
But it would be probably hard to implement. :D

Spiritreader:
> oh wow! thanks.
> But it would be probably hard to implement. :D

It is potentially possible using WMI registry monitoring if we investigate this further.

However something very important to note:
If this can really be used to detect night/day dependind on Windows Night Light, it would require a completely new mode that runs without timer, because the underlying theme switch trigger is now an event!

thebigsmileXD:
Working on it on a fork https://github.com/thebigsmileXD/Windows-Auto-Night-Mode/

Spiritreader:
Just a fair bit of info for you:

Armin was very correct in estimating this to be a lot of work if done right.

This cannot be implemented using a timer, because this would mean that there's a check for a registry key update every 60 seconds, which is something we chose to avoid in ADM.
States are only tracked in memory and synchronized once.

We need a new switch source that is not the timer. 
So essentially, what is necessary is to use a change event listener on the regkey. 
As far as I'm aware, this event handler needs to use WMIC queries.
When the TimedSwitch module is disabled, the offset feature would need to reimplemented to work with that. 
For this, a new mechanism has to be created that is able to correctly postpone a theme switch if it was triggered by windows night light. 

Once that is done, postpone compatibility has to be ensured with all modules, for example the GPU monitoring module.

Finally, there has to be some sort of management class which is able to switch between sources (timer based, event based without restarting the service.

All while doing that on the very dev-active beta 5 branch which currently undergoes heavy restructuring 

There's a good chance I forgot some stuff, but yeah, a quite ambitious project.

Spiritreader:
The Auto Dark Mode Version X structure is now final and has been merged into master.
This version will release on Oct 5th.

If you're still interested in getting this implemented in a future release by doing it "the right way", feel free to reach out.
I'll gladly provide more details explanation on how the inner structure works.

Spiritreader:
Is there still interest for this feature @thebigsmileXD (or anyone else who wants this as a matter of fact) 
If so, we now have everything ready to get this working.

Spiritreader:
[...]

PR #257, opened 2021-09-12
Add bluefilter / night light detection module

thebigsmileXD:
See: https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/issues/256

Theoretically, this works fine. But the issues mentioned in https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/issues/256#issuecomment-916678841 did indeed apply. The time based module still seems to run and it keeps reverting the theme, but i am not sure if that might be caused by the service of the original program still running on my PC, i could not properly debug it yet.

A registryobserver would be a better implementation, currently it is getting changed the same way the timed mode detects it - using a task.

I am still opening this PR so it can be integrated in the future, but with better implementation


Spiritreader:
Thanks for the contribution, i used the bluelight detection mechanism you provided to implement this in #[84f79d6](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/commit/84f79d6f2d3e440b6e02df3d874156c97239cd8d)
along with all the other requirements we talked about.

Closing this now as further issue tracking will be done via github issues

PR #375, opened 2021-11-20
[Bugfix] Reload autostart button getting disabled when disabling autostart entry in Task Manager

janek49:
Fixes the "reload autostart" button getting disabled when disabling the autostart entry in Task Manager. In order to refresh the status info you had to restart the app. Now after reenabling entry in Task Manager you can just click the button to update information.


Spiritreader:
Disabling the autostart refresh button when the entry is disabled in the task manager is intended behavior.
Enabling this in the frontend will allow users to click the button, but it will not reset the autostart entries correctly.

Since disabling the autostart entry is a user-based conscious decision, force-resetting it via auto-dark mode is probably not a good idea.
The validation should only ensure that an entry is set and valid (as in, correct path), but not modify other potential configurations done by the user.

As such, before merging this, we need to define how exactly the validation function should overwrite user modifications to the entries, which can in theory be valid still.

janek49:
I have tested it and it doesn't reenable the disabled task manager entry just polls the status. But if it causes unwanted side effects, it's pointless because adding another button looks bad and won't have any significant userbase, in which case just reject the PR

Spiritreader:
Yes exactly. It doesn't re-enable it. 
But when pressing the button, the validation mechanism should do that in order to be consistent with the intended function of getting autostart back to a functioning state.

I don't think the PR should be rejected based on that, but more that either a related function needs to be updated, or its intended meaning should change.

Which leads to the question: Should we allow users to disable the Auto Dark Mode autostart entry via task manager, or should we only allow disabling it via the app itself? @Armin2208

janek49:
How about a change to the reload button so when the task manager entry is disabled it only rechecks the state without side effects?

Armin2208:
> Should we continue to allow users to disable the Auto Dark Mode autostart entry via task manager, or should we only allow disabling it via the app itself

We shouldn't interfer with Microsofts implementation of Autostart. Task Manager and Settings are the main spots to disable apps starting with Windows. A decision made by the user, which we should respect. Especially on machines with Windows 11. With the latest version of Windows the system always shows a push notifications, as soon as a new app sets a new autostart entry.

Sure, sometimes users can accidentally disable something. Or a third party app can do this. But we should assume that this will never happen.

One thing that is a key factor, is to always have a working autostart entry. The path should be always up to date. Sometime the user will re-enable Auto Dark Mode from the place he disabled it. This should work like expected.

Though if the autostart of Auto Dark Mode isn't enabled, there shouldn't be any updates and no path changes since the last launch. Because no running background service means no running updater.

Spiritreader:
Hmm, in that case I do indeed believe that we should not change the existing behavior.
Even moreso, we should probably go the other route and disable all autostart control in the settings if the entry has been disabled via windows.

Armin2208:
> Hmm, in that case I do indeed believe that we should not change the existing behavior. Even moreso, we should probably go the other route and disable all autostart control in the settings if the entry has been disabled via windows.

UI should adapt to this, yes.
And showing a red colored message: you disabled autostart in windows settings.
With an added link to ms-settings:startupapps

Spiritreader:
> > Hmm, in that case I do indeed believe that we should not change the existing behavior. Even moreso, we should probably go the other route and disable all autostart control in the settings if the entry has been disabled via windows.
> 
> UI should adapt to this, yes. And showing a red colored message: you disabled autostart in windows settings. With an added link to ms-settings:startupapps

we already have that, except for the link
grafik

janek49:
If your going to add a link, maybe we can incorporate a tiny little reload button next to this link to re-check status as I intended with this PR?

Spiritreader:
I've solved the reload problem using WMIC.
There's no need to reload any more in future builds, it will automatically update

janek49:
Just checked it out works great, even better without having to press a button!

PR #403, opened 2021-12-17
Replace ResXFileCodeGenerator with VocaDb.ResXFileCodeGenerator

ycanardeau:
[VocaDb.ResXFileCodeGenerator](https://github.com/VocaDB/ResXFileCodeGenerator) is a [source generator](https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview) that can automatically generate [resource classes](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/857de13ed7482c26f18767aff436cd123a61c4e8/AutoDarkModeConfig/Properties/Resources.Designer.cs) from [resx files](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/tree/857de13ed7482c26f18767aff436cd123a61c4e8/AutoDarkModeConfig/Properties) at compile time. If you don't want to include the `Resources.Designer.cs` file into the [git history](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/commits/a826c4bb30a0226ae5448f1b686aa98b112a4f09/AutoDarkModeApp/Properties/Resources.Designer.cs), then you may find it useful. Although it's not fully compatible with the standard ResXFileCodeGenerator, it seems that there's no problem. (I've tested it on my local machine.) Please feel free to close this PR if this change is not needed.


AraHaan:
No need to use that file code generator, there is a better one known as ``Microsoft.CodeAnalysis.ResxSourceGenerator`` however yes it sadly cannot produce code that is meant to load things like icons, etc as non-string resources.

Spiritreader:
Thanks for the contribution and idea!
I think for now we'll stick with the built-in resource generator though.

We have no problem with including the designer class for now, but are looking into providing a better translator update experience that doesn't involve having to edit xml files.

PR #486, opened 2022-07-14
Update Resources.tr.resx

cmhrky:


Armin2208:
you just removed the years of office. thats not a good solution. Either you still have the years standing their or you add a (legacy) after office.

Armin2208:
great, thanks for correcting that

PR #535, opened 2022-10-06
Update Resources.tr.resx

cmhrky:


Spiritreader (approved):
You are very quick, thanks!

PR #546, opened 2022-10-10
Fix typo

tooomm:
⌚


Spiritreader (commented):
oh it should actually be 3 hours, but good catch

Spiritreader (approved):

tooomm:
In that case...
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/3651872f26c21b0475c6ad4d5129f5a8861472a3/AutoDarkModeSvc/Handlers/ToastHandler.cs#L412-L415
Probably needs an update as well I guess?

Spiritreader:
god damn it :D, will fix. Thanks!

PR #555, opened 2022-10-15
Update Resources.tr.resx

cmhrky:
Thanks for new translations.

This section exists in the translation files but does not appear.
image


Spiritreader (approved):

Spiritreader:
> This section exists in the translation files but does not appear.
That's on purpose as of yet.

Thanks for the update!

PR #559, opened 2022-10-16
Clarify geo coordinates action label

tooomm:
Current wording makes it appear as if it's just a help text explaining what to put into the lat/long fields.
But it's actually a helpful, clickable action that links to a very handy webpage.


Spiritreader (approved):
Make sense, thanks!

PR #577, opened 2022-10-23
Refactor the About Page, both XAML and CS.

PtrStruct:
I've done some refactoring so that the page is easier to work with long term. 
I was constantly finding myself looking all over the place when trying to find things so I figured I'd do some refactoring.

The functionality and the visual appearance is still the same.

- Refactored the XAML in the About Page
- Refactored the CS behind the About Page

Something should be done with Line 424 in the CS file, I can implement it in the next PR if needed?


Spiritreader (approved):
Sure, why not.
I personally don't see the reason behind re-ordering the cs file, but I'm also indifferent about it.

Thanks for cleaning up the remaining grid declarations in the xaml file and also for pointing out that there's a missing error message box call. I'll just quickly push that myself because it's quite trivial.

PR #580, opened 2022-10-24
Refactored the XAML and the CS to follow the same code structure

PtrStruct:
Did some refactoring to follow the same code structure as the About Page.
It's much more pleasant and easy to work with now.

- Refactored the XAML.
- Refactored the CS.

### DIFF
--- a/AutoDarkModeApp/Pages/PageSettings.xaml
+++ b/AutoDarkModeApp/Pages/PageSettings.xaml
@@ -3,9 +3,7 @@
       xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
       xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
       xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-      xmlns:local="clr-namespace:AutoDarkModeApp.Pages"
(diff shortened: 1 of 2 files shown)


Spiritreader:
Hmm, I don't really see any meaningful improvementon the refactor on that one, sorry :(

In general I would recommend inquiring beforehand if a refactor just moves around methods in its order and does some auto formatting.

If you would like to contribute to the UI some more, here's a good starting point:

What the settings page .cs file would need is some logical restructure/refactor that unifies some duplicate logic. 
Also the settings page (and all other ones in fact minus the personalizatin one) require changes to conform a bit more to the W11 style, but we don't have enough resources to refresh it right now)

For example
- using the ErrorMessageBoxes class to display Error Messages vis `ShowErrorMessage(..)` and moving the tray icon enable/disable msg box to the ErrorMessageBoxes class with the option to freely set the text
- adding localization strings for the remaining strings in the .cs file
- putting a border around each section setting excluding the header Settings, Language, Updates, Config) to better match the Win11 setting 
imagesimilarily to how the personalization page is set up: 
image Ideally we would have a new XAML control that wraps it instead of creating a border for each one.

PR #588, opened 2022-10-29
Update Resources.tr.resx

cmhrky:
This line does not exist:
"A downgrade is avaliable"


Spiritreader:
This line does not exist:
"A downgrade is avaliable"

fixed, thanks!

I'll keep the PR open so you can add the missing line if you like

Spiritreader (approved):
Thanks!

PR #598, opened 2022-11-05
Updated Chinese translation for new functions

weaponmasterjax:


Spiritreader:
you deleted the branch before we could take a look at merging.
Also I looked into crowdin and the free tier barely has enough translation strings for us before they expect us to pay.
We would run into that limit very quickly.
So sadly not an option.

PR #618, opened 2022-11-09
Fix download anchor

LitoMore:


Spiritreader (approved):
good catch, thank you!

PR #645, opened 2022-12-01
Added new translations file Resources.resx (Romanian)

emanuel-c4:
I added the Romanian language, maybe some lines are not translated, if there is any missing translation I will correct it.


Spiritreader:
Thank you! Very much appreciated!!

PR #649, opened 2022-12-12
Translated using Weblate (Japanese)

maboroshin:
Last commits were summarized to delete a commit comment. Delete personal information in commit comments.

Today, an option has been added to Weblate to make this private. Instead of merging this PL, You can also retrieve translations again from Weblate.


Spiritreader:
Personal information?
I'm not exactly sure what we should make private.

maboroshin:
Thanks!

Wealate has been embedding email addresses in commit comments. Some repositories do.

I knew this and was using a dedicated Weblate email address. But I still want to keep it private.

On the other hand. The initiative to not embed email addresses seems to have been completed yesterday. (See: #8451 "Allow instance owner to set a default template for commit e-mail, hiding user e-mails from commits by default")

PR #658, opened 2022-12-31
Add Github Action to build on multiple platforms (x86,x64,arm64)

AquaBx:
Hi, I just modified the ``publish.bat`` file to choose the architecture of the build.
I also add a GitHub Action to build automatically on these two architectures.

image

### DIFF
--- a/AutoDarkModeLib/Properties/Resources.cs.resx
+++ b/AutoDarkModeLib/Properties/Resources.cs.resx
@@ -1216,4 +1216,10 @@ On rare occasions, Windows may emit a beeping sound when the mouse is moved duri
   <data name="CursorsPageOpenPointerSettingsDescription" xml:space="preserve">
     <value>You can define new cursor schemes in the Windows settings and then select them here afterwards.</value>
   </data>
+  <data name="AboutHyperLinkOpenShell" xml:space="preserve">
+    <value>Open shell</value>
+  </data>
+  <data name="AboutHyperLinkOpenUpdaterLog" xml:space="preserve">
+    <value>Open updater log</value>
+  </data>
 </root>
--- a/AutoDarkModeLib/Properties/Resources.de.resx
+++ b/AutoDarkModeLib/Properties/Resources.de.resx
@@ -1217,4 +1217,10 @@ In einigen Fällen kann ein Piepston auftauchen, wenn die Maus während der DWM 
   <data name="CursorsPageOpenPointerSettingsDescription" xml:space="preserve">
     <value>Mauszeiger Schemas können in den Windows Einstellungen erstellt, angepasst und dann hier ausgewählt werden.</value>
   </data>
+  <data name="AboutHyperLinkOpenShell" xml:space="preserve">
+    <value>Open shell</value>
+  </data>
+  <data name="AboutHyperLinkOpenUpdaterLog" xml:space="preserve">
+    <value>Open updater log</value>
+  </data>
 </root>
--- a/AutoDarkModeLib/Properties/Resources.el-gr.resx
+++ b/AutoDarkModeLib/Properties/Resources.el-gr.resx
@@ -563,7 +563,7 @@
     <value>Απόκρυψη εικονιδίου περιοχής ειδοποήσεων (δεν συνιστάται)</value>
   </data>
   <data name="tbFeatureDisabledWhileThemeMode" xml:space="preserve">
-    <value>This feature is disabled while Windows manages your theme.</value>
+    <value>Αυτή η δυνατότητα είναι απενεργοποιημένη ενώ τα Windows διαχειρίζονται το θέμα σας.</value>
   </data>
   <data name="tbPickTheme" xml:space="preserve">
     <value>Επιλέξτε ένα διαφορετικό θέμα των Windows για σκοτεινή και φωτεινή λειτουργία αντίστοιχα. Επιτρέπει την πλήρη προσαρμογή του θέματος, αλλά απαιτεί από εσάς να ρυθμίσετε μόνοι σας τα αρχεία .theme.</value>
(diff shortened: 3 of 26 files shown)


Spiritreader:
Hi, thanks for the PR!

I'm not quite sure if we can use your proposed system with our current way of pushing updates. 
Some form of automation would be awesome to have eventually, but we have a few requirements where, without them, it would actually increase our workload instead of lowering it. 
Which is why we have not gotten around to doing this.

I'll list them down below and in case you're interested in pursuing this, feel free to let us know and ask questions if you have them.

### First of all to address the issues with the ARM64 builds:
- Does this also correctly build the rust updater correctly? Does it work?
- When targetting ARM64, is adding the Windows App SDK (which we plan to use in the future) blocking the project from building? We had issues with WinForms and WPF targetting ARM64 while using the AppSDK in the past
- Our update system is not at all compatible with distributing different archs. Before we support that, there need to be breaking changes made to the entire update system. That is a showstopper for now for ARM builds.

### Then to address issues with using github actions atm
- New builds need to be triggered when the **build number** which can be found here: https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/ff680493f5994c135f4df410fb2af06807e17d2d/AutoDarkModeSvc/AutoDarkModeSvc.csproj#L8 updates, instead of a blanket solution like on push or on a new merge PR.
- The Action would need to run in our beta version repository, or at least, push the artifacts as a new build to that repo: https://github.com/AutoDarkMode/AutoDarkModeVersion
  - Create a new preview release in that repo for each new artifact 
  - Update the version XML file here: https://github.com/AutoDarkMode/AutoDarkModeVersion/blob/master/version-beta.yam to contain the correct version and links to the download files.

AquaBx:
For now, I modified the yaml script to build automatically on new version. 

For the arm64 case, well I build a dozen times the program and it runs pretty well on my Surface Pro X.
I don't saw any particular error during the build, but maybe i miss something during the configuration.
I will do issues if i found any bugs on my Surface.

AquaBx:
If you want to see the logs of the last build 

https://github.com/AquaBx/Windows-Auto-Night-Mode/actions/runs/3820216768/jobs/6498364174

Spiritreader:
> For the arm64 case, well I built a dozen times the program and it runs pretty well on my Surface Pro X.

Can you check whether there are exceptions found in the ADM log file with the "Restore themes when changed externally" flag enabled here?
image

Unfortunately there's still no support for the arin10-arm runtime identifier with WinAppSDK while using WPF or WinForms:
image

AquaBx:
I get this error when checking this parameter in the service.log file

2023-01-02 15:03:01 | Error | WindowsThemeMonitor.StartThemeMonitor: could not start active theme monitor TypeInitializationException > The type initializer for 'System.Management.M
anagementPath' threw an exception. > System.Management.ManagementPath get_DefaultPath()

[service.log](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/files/10332017/service.log)

Spiritreader:
> 



> I get this error when checking this parameter in the service.log file
> 
> ```
> 2023-01-02 15:03:01 | Error | WindowsThemeMonitor.StartThemeMonitor: could not start active theme monitor TypeInitializationException > The type initializer for 'System.Management.ManagementPath' threw an exception. > System.Management.ManagementPath get_DefaultPath()
> ```
> 
> [service.log](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/files/10332017/service.log)

Yep so far that only happens when building for ARM iirc. The full exception message is that some components of System.Management are currently only supported for x64.

Additionally, in order to user the WindowsAppSDK, we have to put a runtime identifier along with the self-contain options of WinAppSDK
```
    <RuntimeIdentifiers>win10-x64</RuntimeIdentifiers>
    <WindowsPackageType>None</WindowsPackageType>
    <WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained>
```
Which limits it to x64, otherwise the project won't compile.

AutoDarkModeSvc/AutoDarkModeSvc.csproj
```diff
@@ -19,6 +19,10 @@
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <SupportedOSPlatformVersion>10.0.18362.0</SupportedOSPlatformVersion>
+
+    <RuntimeIdentifiers>win10-x64;win10-arm64</RuntimeIdentifiers>
+    <WindowsPackageType>None</WindowsPackageType>
+    <WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained>
```
AquaBx, line 25: Well i added that here but I'm not shure, because i tried without the `RuntimeIdentifiers``and it compile perfectly fine.

AquaBx: ok my bad, i don't know why it remove the package

AquaBx:
well when i try this `<RuntimeIdentifiers>win10-x64;win10-arm64</RuntimeIdentifiers>` i got another error and for now I don't the level to understand why

AquaBx:
by adding  `/p:Platform=%arch%` on all dotnet command in `publish.bat` i am able to build without 
` <RuntimeIdentifiers>win10-x64</RuntimeIdentifiers>`

however there is a lot of warnings.

AquaBx:
caused by TargetPlatformMinVersion

i changed it to 10.0.18362.1 for tests

AquaBx:
> caused by TargetPlatformMinVersion
> 
> i changed it to 10.0.18362.1 for tests

it changed nothing on github action, but locally yes 🤷

AquaBx:
since 03/15/2022 and the 1.0.1 version , Microsoft provide a arm64 
https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/downloads#current-releases
image

AquaBx:
Well i tried to compile the new version, and it tells me that 'publish-inject-updater-legacy-files.bat' is missing

AquaBx:
I updated my github action to be able to compile the new version. It seems to work.

AquaBx:
But as you mentionned ManagementEventWatchers  is still a problem as it throws errors :

image

Spiritreader:
Yup, if that gets resolved we should be able to start testing on arm.

PR #664, opened 2023-01-11
Add access keys to tray menu

Susko3:
[Access keys](https://learn.microsoft.com/en-us/windows/apps/design/input/access-keys) allow quickly selecting an option in a menu using only the keyboard. Access keys are defined with an ampersand `&` just before the desired character.

For my use case, this allows very quick access to toggling the theme and forcing light or dark theme. Previously, I would have to right click on the tray icon, and then snipe the buttons with my mouse. Access keys allow me to right click, and then press the desired key: <kbd>T</kbd> for `Toggle theme`, <kbd>L</kbd> for `Force Light Theme`, etc.

For me, using access keys is much faster, and adding access keys does not impede existing point-and-click usage.

How the access keys display (notice the underlined characters):

image

Please double-check that I've done the translations correctly, I've never used the translation system you use before.
Would it be possible to tag the copied translations as 'needs retranslation'? Or is that something that will happen automatically?

### DIFF
--- a/AutoDarkModeLib/Properties/Resources.Designer.cs
+++ b/AutoDarkModeLib/Properties/Resources.Designer.cs
@@ -2556,7 +2556,7 @@ namespace AutoDarkModeLib.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Pause auto switch.
+        ///   Looks up a localized string similar to &amp;Pause auto switch.
         /// </summary>
         public static string TrayMenuItemThemeSwitchPause {
             get {
--- a/AutoDarkModeLib/Properties/Resources.Designer.cs
+++ b/AutoDarkModeLib/Properties/Resources.Designer.cs
@@ -2520,7 +2520,7 @@ namespace AutoDarkModeLib.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Automatic theme switch.
+        ///   Looks up a localized string similar to &amp;Automatic theme switch.
         /// </summary>
         public static string TrayMenuItemAutomaticThemeSwitch {
             get {
@@ -2529,7 +2529,7 @@ namespace AutoDarkModeLib.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Close.
+        ///   Looks up a localized string similar to &amp;Close.
         /// </summary>
         public static string TrayMenuItemClose {
             get {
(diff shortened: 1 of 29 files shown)


Spiritreader:
That's a really cool addition.
I'll take a look into merging asap!

Susko3:
Keep in mind you might need to enable 'Underline access keys when available' in Windows [accessibility settings](ms-settings:easeofaccess-keyboard) to see the access keys. Or access the menu using only the keyboard: with a combination of <kbd>Win</kbd>+<kbd>B</kbd> and then <kbd>Shift</kbd>+<kbd>F10</kbd> to open the tray menu of the focused icon.

Spiritreader:
> Keep in mind you might need to enable 'Underline access keys when available' in Windows keyboard accessibility settings (found at `ms-settings:easeofaccess-keyboard`) to see the access keys. Or access the menu using only the keyboard: with a combination of Win+B and then Shift+F10 to open the tray menu of the focused icon.

Awesome, thanks.
Would it also work if we had for example a hotkey that opens the tray menu?

Susko3:
> Would it also work if we had for example a hotkey that opens the tray menu?

I think it should show the access keys, but I don't know how to test that.

> Also if accessibility keys show by default when holding alt, it may be possible to remove alt from the key listener such that they can be triggered even if it's off.

<kbd>Alt</kbd> is only useful for when the access keys are in a toolbar / menu bar (tapping <kbd>Alt</kbd> will focus the toolbar and show the access keys). In my testing, pressing alt when a tray menu is open will close it (in Auto Dark Mode and all other tray programs).

Also the access keys can be used even if they are not shown/underlined.

Example menu bar:

image

Spiritreader:
> > Would it also work if we had for example a hotkey that opens the tray menu?
> 
> I think it should show the access keys, but I don't know how to test that.
> 
> > Also if accessibility keys show by default when holding alt, it may be possible to remove alt from the key listener such that they can be triggered even if it's off.
> 
> Alt is only useful for when the access keys are in a toolbar / menu bar (tapping Alt will focus the toolbar and show the access keys). In my testing, pressing alt when a tray menu is open will close it (in Auto Dark Mode and all other tray programs).
> 
> Also the access keys can be used even if they are not shown/underlined.
> 
> Example menu bar:
> 
> image

Ah alright I see, thanks for the info.
I've found a way to always show access keys via userdll32. I'll add that with the merge.

Spiritreader (approved):

PR #680, opened 2023-01-26
Update Resources.uk.resx

MykhaiIo:
Fixed tray icon strings, and reversed word order in strings followed by the hotkey string (still not natural due to the predefined location of the 'hotkey' word, but slightly better than before), and other minor corrections. 
Looks like there are still no strings for custom scripts documentation and repository available for translation.


Spiritreader (approved):
Hey, thanks!

Yes, that's correct, there are none yet for the custom scripts section as that is a WIP.

Issue #733, opened 2023-06-04
Add option to disable theme switching while some process is running.


pohuing:
**Describe the enhancement or feature you'd like**
Option to disable switching if certain programs are running.

**Describe alternatives you've considered**
The idle switch and the low GPU usage switches. 

**Additional context**
The low GPU usage threshold does not work on my AMD GPU, because it's either too beefy, or the sampling is just broken(AMD reports null for sensors occasionally, so this isn't an issue with this application itself).
Furthermore, the GPU load threshold is inadequate for some scenarios:
I don't mind if my system switches while watching videos, or doing a render, or streaming per se. In those situations the frame drops are irrelevant, yet would be prevented by GPU usage. 
On the other side, plenty of timing specific games use very little resources, such as League of Legends, Counter Strike, any roguelike really, similar for a lot of fighting games like Rivals of Aether.

Armin2208:
Idea: don't switch while Game Mode is activated as an additional option for the "Don't switch while playing video games" setting.

pohuing:
I don't know anyone who uses Game Mode to be honest. Having a list of processes to block the switching sounds the most straight forward to me.
Also, I'm not aware of an API to actually check if it's active. The docs are pretty lackluster and the only somewhat related API I can find has been deprecated for years now.

pohuing:
I'll take a shot at implementing this myself. I'm not very familiar with WPF and C# anymore though.

Armin2208:
Every gamer uses Windows Game Mode, because it's enabled by default for every game. People need to opt out of Game Mode to have it disabled.

But you are right, listening if game mode is activate is not good documented. PowerToys uses a method in Always on Top.
![grafik](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/assets/32982430/6eee864c-0986-4e51-84ea-fcfe0860b660)

Btw thanks for your pull request!

Spiritreader:
Your implementation is now available in the latest beta.

pohuing:
Was fun working on this, thanks for the quick help!

PR #734, opened 2023-06-05
Add feature to disable theme switching while certain processes are running

pohuing:
This is a rough fix for #733 currently it can
- show a list of processes
- add a process to a block list 
- check for active processes to postpone switching

I'm opening this PR early before I'm done because I'm a bit lost with regards to the following topics.

What's missing as far as I can tell is
- localization 
  - This would happen eventually via Weblate?
- A better display for excluded processes
  - Currently it uses a DataGrid, I would love something that looks similar to the Windows' settings
image

  - Not sure how to implement that however
- An option to remove entries
  - Related to the prior point

What I'm not sure about is the current way to display a list of active processes. I filter for processes with an active handle and a window. This rules out about 100 instances of svchost, but not everything I wouldn't consider an application the user interacts with. For instance ApplicationFrameHost, svchost, and TextInputHost below:
image
Furthermore I currently require the user to refresh a process list because 
- I'm not sure how expensive getting the entire process list is
- Automatically reordering lists while the user is interacting with them is terrible UX
maybe there's a better solution I'm not aware of.

Then there's the behavior of the ProcessBlockModule. Running the AutoDarkMode service seems to switch themes immediately, without waiting for the module to be initialized. If I instead start the application and have it wait for a custom time in the near future my Module seems to work fine.

I would appreciate it if someone more experience than me with .NET and WPF could help me out here. Thanks in advance!


pohuing:
I've figured out a way to display the block list in a prettier manner and added the option to remove entries. It's not super pretty but my lack of taste in design fits my profession 😆. The thing now looks like this:
image
Opening the ComboBox now refreshes the list of active processes, this causes some noticeable lag for about a quarter of a second, but I think is overall easier to use than having to manually refresh.
"Entfernen" is German for Remove. I believe my localization works for German and English systems.

I'm still unsure if my service module works correctly or not, I'd be more comfortable if someone could verify that before I'd consider this ready to merge. Are there tests I could expand to cover my code?

Spiritreader:
Hey, thanks for working on this!
Unfortunately I'm not a WPF developer either, I hacked in the current look of the UI.

I think going for a function first approach similar to how you did it is totally fine for now.

If you have any questions about the backend, feel free to reach out.
I'll probably be able to get back to you tomorrow because of the holiday.

>  Running the AutoDarkMode service seems to switch themes immediately, without waiting for the module to be initialized. If I instead start the application and have it wait for a custom time in the near future my Module seems to work fine.

That is somewhat of a design limit currently because a theme switch request is fired immediately after the application starts.
You don't have to worry about that, I'll look into this at some point.

Spiritreader (changes requested):

AutoDarkModeSvc/Modules/ProcessBlockList.cs
```diff
@@ -0,0 +1,57 @@
...
+            return;
+        }
+
+        var activeProcesses = Process
+            .GetProcesses()
+            .Where(p => p.MainWindowHandle != 0)
+            .Select(p => p.ProcessName)
+            .ToHashSet();
+
+        if (ConfigBuilder.Config.ProcessBlockList.ProcessNames.Any(p => activeProcesses.Contains(p)))
```
Spiritreader, line 39: Detection logic shouldn't run constantly.

Similar to the GPUMonitorModule, it should start monitoring a minute or two (for the GPU monitor module that value is customizable) before a theme switch would occur, then hold and monitor for as long as necessary. After the program that causes the delay is closed, the module should sleep again and not perform any costly operations.

See here for more details:
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/22f76896cae74b52d00ea0c7edc9879a93066eca/AutoDarkModeSvc/Modules/GPUMonitorModule.cs#LL69C35-L69C35

This is fhe function that is used in the above mentioned snippet.
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/22f76896cae74b52d00ea0c7edc9879a93066eca/AutoDarkModeSvc/Modules/GPUMonitorModule.cs#L230

Spiritreader (changes requested):

AutoDarkModeSvc/Modules/ProcessBlockList.cs
pohuing: So in principle my conditions for running my process tests would be
- Is the current time shortly before the scheduled switch
- Is the current theme different from the intended theme (for example after resuming the PC from suspend)

Is there a property I can check for that? I'll be honest the mechanism of the GPUMonitorModule is not very understandable to me. I suppose I could just copy the Fire from that and replace the check with my own, but as far as I can tell it also doesn't check for such irregular switches.

Spiritreader: > Is there a property I can check for that?

That's a very good point!
There is not at the moment, however there should be one, as that allows for simpler module design and absolves the dev implementing a module of the responsibility of handling when exactly modules may run.

> Is the current theme different from the intended theme (for example after resuming the PC from suspend)
Modules should only postpone user and time/nightlight-based switches, never irregular or event-based ones (such as resuming from standby or battery-based switches). 
As such it should only be necessary to activate the module when the new property indicates that modules should start operating.

____________________

I will implement such a property, which should mean that you don't have to worry about any of this. You will merely need to implement checking whether that property returns a value that indicates modules are allowed to start their postpone logic once that's done.

Spiritreader:
I've added a ThemeSwitchApproaching proprty to the globalstate which you can use do determine whether a module is allowed to start running. There's an example in the GPUMonitorModule
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/9aa1ef6211c3974a5722b4cb6b6a83b3a00cab8e/AutoDarkModeSvc/Modules/GPUMonitorModule.cs#L59

Should be a lot more streamlined now and is governor independent.

Spiritreader:
In addition, there's also the necessity to register modules that hook into that mechanism via the module's enable hook

https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/e58ed97c4e315af44aaa2ea4a5443b2e12dcfe04/AutoDarkModeSvc/Modules/GPUMonitorModule.cs#LL177C12-L177C12

And the disable hook
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/e58ed97c4e315af44aaa2ea4a5443b2e12dcfe04/AutoDarkModeSvc/Modules/GPUMonitorModule.cs#L182

pohuing:
I believe this feature now works as intended. Or at least it works on my end and doesn't seem to do unnecessary process checks. 
I decided not include my own monitoring flag, instead just checking if I currently postpone based on the postpone reason. 

I think that should be fine since I don't expect there to be a whole lot of postpones in the queue.

pohuing:
The filtering for the dropdown is still a bit weird, listing processes I wouldn't consider relevant for the feature. I'll look around how other applications handle it but this is a cosmetic issue and does not impact functionality nor usability significantly.

AutoDarkModeApp/Pages/PageSwitchModes.xaml
```diff
@@ -109,11 +109,90 @@
...
+                        <DockPanel Margin="0,10,0,0">
+                            <Expander
+                                DockPanel.Dock="Bottom"
+                                FlowDirection="RightToLeft"
+                                Header="{x:Static p:Resources.SwitchModesExpanderHeaderBlockedProcesses}"
+                                HorizontalAlignment="Stretch"
+                                Margin="0,10,0,10">
+                                <Expander.HeaderTemplate>
+                                    <DataTemplate>
+                                        <!--Unsure why the negative left margin is necessary here,
```
Spiritreader, line 134: This is because the WPF header was never designed for somethign like this.
I actually had to completely overwrite the header template on the accent color page to get this to work, so I don't think there's an easy way to do this. Hence I think it's fine for now.

Spiritreader:
> I decided not include my own monitoring flag, instead just checking if I currently postpone based on the postpone reason.
> I think that should be fine since I don't expect there to be a whole lot of postpones in the queue.

That is correct. I usually use the monitoring flag because it's easier to read, but using the postpone queue is also absolutely fine.

> I believe this feature now works as intended.

Awesome!

PR #783, opened 2023-10-05
ci: Add Dependabot

ottnorml:


Spiritreader:
Hi, thanks for setting this up!

As far as I can see you've only set this to minor version changes and patches. This is actually what we want, just making sure I read this right.

DinuruSeniya:
Any update on this PR? @Spiritreader

PR #882, opened 2024-10-06
Update PageSettings.xaml

olegtaratuhin:
Extra margin in checkboxes in Settings Page Updates section might have implied that the options are nested, while they are not.


Spiritreader (approved):
good call, thanks!

Issue #925, opened 2025-03-17
Request: About use WinUI 3 to rebuild


ChenYiLins:
Using WinUI to refactor a program is a difficult thing, but it will make many operations easier. I believe that WinUI has now reached a relatively mature state. I started the work on an accidental idea. Based on the source code, I completed 66.6% of the page construction. But its workload is actually far beyond my imagination, because upgrading to Windows App Sdk makes many works easier, but  some of the related code needs to be rewritten. At the same time, some codes become confused because of long-term updates.

Using WinUI to refactor the program may solve these issues: #895 #874 #702 and more

I hope I made a good start for others, more people can participate in the construction of the code, and it will be more helpful to get a complete Auto Dark Mode built by WinUI as soon as possible.

Maybe manager can create a new branch for its development. I put the current project in this [repositorie](https://github.com/ChenYiLins/Windows-Auto-Night-Mode). More information could see in _.github/DEVELOPMENT_NOTE.md_. There may be some questions not mentioned in the document, so you can ask them at any time under this issuse.

A few extra words: At the same time, I have improved the localization of English, French, Japanese and Simplified Chinese, which is believed to be helpful for many developers to test. However, the localized text can't be seamlessly switched to the new code now, so you can refer to the development documentation to make a little modification.

The following is the unfinished Page: 
- WallpaperPicker
- Colorization
- Cusors
- ThemePicker

Spiritreader:
We're aware and you're right.

But there's neither time nor resources to do a complete overhaul. We'd need more contributors, simply speaking.

ChenYiLins:
Thanks for your reply. Indeed, it requires a lot of resources and time. As you said, it is a good idea to create a new branch to deal with it. 
( ꒰｡ › . ‹ ｡꒱ )

Jay-o-Way:
Basically #895, but much better explained 😆

Jay-o-Way:
@ChenYiLins much respect for starting on this! My personal opinion is that it is good to start from nothing. Just to prevent any old code from being copied. I know it's a huge work, but I like the fresh-ness. Keeping new code separate is also a great idea. Also think that - by the time the new version is ready - we should call it v2.

Few suggestions:

- I agree using `x:uid` with _resw_ looks like a good idea.
- Check out the new XAML TitleBar control in wasdk 1.7 ✨
[...]

PR #929, opened 2025-03-28
Initial commit

ChenYiLins:
- This submission is related to: #925
- Components that are not used at present have been deleted. If necessary, please add them directly from the master branch to the current branch.


Spiritreader (dismissed):
You've configured your code formatter to expand short hand get setters

```c#
public AdmConfig Config { get; private set; }
```

to

```c#
public AdmConfig Config
{
    get; private set;
}
```

~~Could you revert this change?~~ I've adjusted the formatter settings.
I don't really mind if this is done in the AutoDarkModeApp, but for the other components I would prefer it to remain unchanged.

Spiritreader:
> Components that are not used at present have been deleted. If necessary, please add them directly from the master branch to the current branch.

Also, merging this initial commit should only be done once all non UI components have been re-added in their previous (or future agreed upon) state

ChenYiLins:
> > Components that are not used at present have been deleted. If necessary, please add them directly from the master branch to the current branch.
> 
> Also, merging this initial commit should only be done once all non UI components have been re-added in their previous (or future agreed upon) state

You are right, this can avoid many repeated commits.

> I've gone ahead and re-added the missing backend files, 和 separated the winui3 build from the backend for now, such that it's easier to work on it.
> 
> Also the target has been upped to .net9
> 
> In addition, I've adjusted the code formatter rules to match the previous style for single line auto properties and added a few more for namespace formatting.
> 
> Please check if everything still compiles, you might need to manually delete the `obj` directories within the modules and the `Debug` directory.
> 
> So in that regard, the initial PR is ready to merge into the winui3 branch. I will be waiting for further comments.

Thank you for adding.

It is wise to upgrade to .net9, because it can promote the realization of Aot. At the same time, need to make changes to the code of ViewModels.

A unified code style can make the code easier to maintain. If necessary, additional documents can be added to explain.

Jay-o-Way:
Cross-referencing https://github.com/ChenYiLins/Windows-Auto-Night-Mode/pull/1

Jay-o-Way:
@Spiritreader @Armin2208 at what point would you like to pull this in? Then I can use the adm repo to base a next pr on.

ChenYiLins:
> Also, merging this initial commit should only be done once all non UI components have been re-added in their previous (or future agreed upon) state

@Jay-o-Way This counld avoid many seemingly repetitive commit before the basic work is completed.

ChenYiLins:
@Jay-o-Way Do you run in package? Please make sure you are running in unpackage. Maybe I should delete it.

Spiritreader:
> Now I get this :( image

Yes, it seems like you've been running the packaged (msix) version so far.
Make sure the build is on unpackaged. Also selecting yes on that message should not break anything.

PR #934, opened 2025-04-01
more UI updates

Jay-o-Way:
- fixed title bar icon (color)
- moved controls around in the Update area - please check!
- moved all styles into one file

### In progress:

+ Remove Office
+ merge MainWindow + ShellPage
+ re-write **Navigation** handling
  - see https://learn.microsoft.com/windows/apps/design/controls/navigationview#navigation


AutoDarkModeApp/Views/TimePage.xaml
```diff
@@ -180,30 +177,21 @@
...
-                                    Minimum="-60"
                                     SmallChange="1"
                                     SpinButtonPlacementMode="Inline"
                                     Value="{x:Bind ViewModel.OffsetDark, Mode=TwoWay}" />
                             </StackPanel>
                         </controls:SettingsCard.Header>
-                        <!--  TODO: remove this button!  -->
-                        <Button
-                            MinWidth="120"
-                            Command="{x:Bind ViewModel.SaveOffsetCommand}"
```
Jay-o-Way, line 205: Note to self: double check

AutoDarkModeApp/Views/SettingsPage.xaml
```diff
@@ -75,31 +75,28 @@
...
-                        <ComboBoxItem Content="{helpers:ResourceString Name=UpdatesComboBoxInterval14}" />
-                    </ComboBox>
-                    <controls:SettingsExpander.Items>
+                        <!--  frequency  -->
+                        <controls:SettingsCard Header="{helpers:ResourceString Name=UpdatesTextBlockUpdateInterval}" IsEnabled="{x:Bind ViewModel.IsUpdaterEnabled, Mode=OneWay}">
+                            <ComboBox SelectedIndex="{x:Bind ViewModel.SelectIndexDaysBetweenUpdateCheck, Mode=TwoWay}">
+                                <ComboBoxItem Content="{helpers:ResourceString Name=UpdatesComboBoxInterval1}" />
+                                <ComboBoxItem Content="{helpers:ResourceString Name=UpdatesComboBoxInterval3}" />
+                                <ComboBoxItem Content="{helpers:ResourceString Name=UpdatesComboBoxInterval7}" />
+                                <ComboBoxItem Content="{helpers:ResourceString Name=UpdatesComboBoxInterval14}" />
```
Jay-o-Way, line 96: I would just use: every day/week/month.

AutoDarkModeApp/Views/SettingsPage.xaml
```diff
@@ -75,31 +75,28 @@
                 </controls:SettingsExpander>
 
                 <!--  Updates  -->
-                <!--  TODO: refactor this UI  -->
```
Jay-o-Way, line 78: Please do check

AutoDarkModeApp/MainWindow.xaml.cs
```diff
@@ -17,16 +17,11 @@ public MainWindow()
         InitializeComponent();
 
         AppWindow.SetIcon(Path.Combine(AppContext.BaseDirectory, "Assets/AutoDarkModeIcon.ico"));
+        AppWindow.SetTaskbarIcon(Path.Combine(AppContext.BaseDirectory, "Assets/AutoDarkModeIcon.ico"));
```
Jay-o-Way, line 20: Check this

AutoDarkModeApp/Views/SettingsPage.xaml
ChenYiLins: Rationalization proposal. This is actually quite suitable for the update frequency of the releases. (Maybe? φ(*￣0￣) )

AutoDarkModeApp/Views/ShellPage.xaml
```diff
@@ -25,7 +26,7 @@
             IsPaneToggleButtonVisible="True"
             PaneToggleRequested="NavViewTitleBar_PaneToggleRequested">
             <TitleBar.IconSource>
-                <BitmapIconSource UriSource="Assets\AutoDarkModeIcon.ico" />
```
ChenYiLins, line 28: Actually, when I didn't look at the code, I always thought that the black and white title bar icon was intentional. (Just like adding Dev text to the title o.O )

AutoDarkModeApp/Views/SettingsPage.xaml
Jay-o-Way: This is a choice that has to be made by the owners. Just throwing it up.

Jay-o-Way:
Off-topic: I'm getting breaking errors. First, a few of these in a row:

> Exception thrown: 'System.TimeoutException' in System.IO.Pipes.dll

Then:

> Exception thrown: 'AutoDarkModeApp.Services.SwitchThemeException' in AutoDarkModeApp.dll
Exception thrown: 'System.Runtime.InteropServices.COMException' in WinRT.Runtime.dll
Exception thrown: 'System.Runtime.InteropServices.COMException' in System.Private.CoreLib.dll
An exception of type 'System.Runtime.InteropServices.COMException' occurred in System.Private.CoreLib.dll but was not handled in user code
Er is een asynchrone bewerking niet correct gestart.

> Only a single ContentDialog can be open at any time.

And CoPilot says:

> This error happens because only a single ContentDialog can be open at any time in a UWP or WinUI application. The exception System.Runtime.InteropServices.COMException with the message "Er is een asynchrone bewerking niet correct gestart" indicates that an attempt was made to show a ContentDialog while another ContentDialog is already being displayed.
This might be happening because:
> 1.	The ShowErrorMessage method is being called multiple times before the previous dialog has been closed.
> 2.	There might be other parts of the application that are also trying to show a ContentDialog simultaneously.
To fix this issue, you can add a check to ensure that no other ContentDialog is open before showing a new one. Here is a possible solution:

Jay-o-Way:
@Armin2208 or @Spiritreader I am currently trying to merge MainWindow + ShellPage, but it's super complicated. I keep finding more Helpers (like TitleBarHelper), Services (like NavigationService, ActivationService), and so on... It's more than I understand. I can upload what I have now, but I need help with this, because every problem that I try to fix leads to another.

Spiritreader:
> @Armin2208 or @Spiritreader I am currently trying to merge MainWindow + ShellPage, but it's super complicated. I keep finding more Helpers (like TitleBarHelper), Services (like NavigationService, ActivationService), and so on... It's more than I understand. I can upload what I have now, but I need help with this, because every problem that I try to fix leads to another.

I am not entirely sure what ShellPage is supposed to be.
Also the NavigationService and ActivationService appear to come from the rewrite done by @ChenYiLins, so maybe they can give an answer on what they are doing precisely.

___________


One possibly reason why you appear to be getting errors here:
> Exception thrown: 'System.TimeoutException' in System.IO.Pipes.dll

may be because the frontend can't seem to connect to the backend. Especially because it originates from a SwitchThemeException.

I would first debug and check where exactly this exception originates from.

Previously, in the old UI, the service would not get started by the frontend if a debug environment is detected.
I don't know how this is handled here as I haven't had time to review exactly what's going on with the rewrite.
Did you start the backend beforehand (AutoDarkModeSvc)?
You most likely need to configure Visual Studio to start both the App and the Service if you want to test functionality in tandem with the backend)

_________

Also @ChenYiLins , there are some oddities in how the frontend now interacts with the messaging client
https://github.com/Jay-o-Way/Windows-Auto-Night-Mode/blob/13ce96e9f4376def0e518366a0c0c4ef1fed2d33/AutoDarkModeApp/ViewModels/TimeViewModel.cs#L248

The pipe client itself should handle all exceptions and wrap them in the ApiResponse.
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/winui3/AutoDarkModeComms/PipeClient.cs

The `SendMessageAndGetReplyAsync` method already wraps the synchronous one, so awaiting it doesn't really make sense. Instead, the synchronous version should be put into the dispatcher queue.

ChenYiLins:
@Jay-o-Way @Spiritreader Comming~ Sorry to have waited so long.

### First of all

For merging MainWindow and ShellPage. I have finished it. Pls see [it](https://github.com/ChenYiLins/Windows-Auto-Night-Mode/commit/908852557546d33f3fee17b2ba8a05a476991a38) (Of course, this branch is just for demonstration now, because the processing reasons of the branch can't be merged directly now, so please copy and paste these codes as much as you like.)

> I am currently trying to merge MainWindow + ShellPage, but it's super complicated. I keep finding more Helpers (like TitleBarHelper), Services (like NavigationService, ActivationService), and so on... It's more than I understand. I can upload what I have now, but I need help with this, because every problem that I try to fix leads to another.

- TitleBarHelper: Help TitleBar update with the Theme update. It is mainly used to solve some mistakes in custom TitleBar, including but not limited to Color of Button. I'm not sure if I need it in WAS1.7.
-  NavigationService: As the name implies, it is used for Nav to better navigate between different pages.
- ActivationService: Run something when the MainWindow is initialized. For example, restore the size and position of the MainWindow.

( To be honest, most of these contents actually come from Template Studio. (っ °Д °; )っ )

I have read your commit, and the mistake is because you have lost the key link of injection, especially the ShellViewModel. So I changed the method of creating the MainWindow and passed in the ShellViewModel.

### Anymore

> I am not entirely sure what ShellPage is supposed to be.
> 
> If you're referring to App.xaml and App.xaml.cs, it's main focus is to start the service if it hasn't started and manage the jump list entires. Previously it was also responsible for launching UI components depending on whether ADM runs on Win10 or Win11, which is why it was separate from the MainWindow.
> 
> It also managed the App mutex which prevents the App from being opened multiple times, sends commands to the backend to set auto start entires etc.
> 
> My first intuition would be that this logic should remain separate from the MainWindow as it is mostly unrelated to the actual UI functionality itself.
> 
> The NavigationService and ActivationService appear to come from the rewrite done by @ChenYiLins, so maybe they can give an answer on what they are doing precisely.

The original intention of ShellPage is only to separate TitleBar and Nav, and there is no other purpose. These functions you mentioned deserve our attention, but they should not be realized by ShellPage, because it only focuses on UI.

At the same time, the UI components are adapted to Win10 and Win11 respectively, which should be well implemented in ViewModel, and there is no need to load these at the beginning. Because MVVM φ(*￣0￣)  .

> `SendMessageAndGetReplyAsync` method already wraps the synchronous one, as such the `async () => await` might not be necessary. If you need a synchronous call, you can use `SendMessageAndGetReply`straight away.
> 
> Also some general information: The pipe client itself should handle all exceptions and wrap them in the ApiResponse object, so it should always be "safe" to call it. The `ApiResponse` then contains the exception and should be unwrapped as needed in the frontend. https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/winui3/AutoDarkModeComms/PipeClient.cs

These codes are almost copied from the old ones, and I didn't even have time to figure out the function of some codes. But your suggestion is very useful. Thank you very much!

Spiritreader:
> I absolutely prefer to use the x:Uid approach.

The considerations that need to be made with this are:

- Localization is also used in tray for the backend.
- The backend (service) must not have any winui3 dependencies in it
- The backend should remain a separate service (not integrated into the UI)

The current resx approach which shares localization strings between service and app was used due to convenience.
We already had the translation strings, and could re-use them. That's all there was to it.

There are a couple of options here, including but not limited to
1. Porting / migrating the entire localization infrastructure (including the weblate translation system) to x:uid directives
2. Canning all translations and re-localizing everything (@Armin2208 and the vast majority of our volunteer translators will be very unhappy with this)
3. Offering the winui3 application as an option to toggle.

Potentially a combination of those points can make sense.

_____________

> All of this, is why I said to start from scratch

This is one of the main reasons why neither @Armin2208 nor I have considered a WinUI3 update, as the functionality of ADM primarily (at least as designed) primarily comes from configuring it once, and then letting the service do its job, with minimal interaction with the UI.
It's about weighing benefit vs effort vs time investment. 
However, it's of course undeniable that using WinUi3 will result in better UX.

When truly starting from scratch, in my opinion, the most sensible way would be to use proper Software Engineering to ensure the right design decisions are taken.

Another really important step is to define requirements, namely what the App's/UI's responsibilities are, and how exactly it interfaces with the service.

There's also the question how much of the current attempt at creating a winui3 app can be reused for a rewrite, it would be sad to discard current progress, but understandable if that structure does not conform to the canonical way to develop winui3 apps.

And, to re-state: I am not a WPF nor a UWP / WinUi dev in the Microsoft ecosystem space.
When it comes to frontend development, I come from TS/Vue and have picked up the ADM WPF app as necessity when becoming a primary maintainer. 

While I can give pointers about the architecture of ADM, I would not consider myself competent at making framework-specific design decisions. Instead, my process of thought targets a more high-level look at the application itself.

Finally, due to the decoupled architecture of ADM, it is in theory possible to create a standalone UI package that has zero dependencies or tie-ins to the service, as long as a messaging client exists which correctly implements the way to talk to the service via named pipes.

Rewriting the frontend may also require a more direct line of communication between contributors than GitHub issues and PRs alone, such that pressing questions can be collected and answered more efficiently.

ChenYiLins:
> Note that the last release for Template Studio seems to be in 2023. I don't think this should be used, at all. Visual Studio can suggest a lot, possibly combined with Copilot. Sorry, but **the helpers really need to go**. And again, I want to say that ~I might not know the current solution for strings (resources) _very_ well, but~ I absolutely prefer to use the **x:Uid** approach.
> 
> All of this, is why I said to start from scratch... WinUI 3, or WASDK 1.7, **is not the same** as a different and older platform. You really can't simply change some properties an try to use what you've already been using for years. Just being honest here...

@Jay-o-Way To be honest, do we have any examples to refer to? I see that [WinUI-Gallery](https://github.com/microsoft/WinUI-Gallery) or [PowerToys](https://github.com/microsoft/PowerToys) are all using ShellPage, including similar services. Now the “cleaning” work has really become a bit difficult.

Jay-o-Way:
I know WinUI Gallery and PowerToys too. The first has a pr open to change exactly this (and also using the new TitleBar control) and PowerToys is honestly a huge mess. It uses just about any framework, it's written by hundreds of volunteers, each having their own style, and the largest part of the code is simply (too) old.

Jay-o-Way:
> I am not a WPF nor a UWP / WinUi dev in the Microsoft ecosystem space

I totally understand! I'm just a simple volunteer too. With my own experience/preference/opinion... Everything I think to know is learned on-the-go, because I wanted to help out with some projects. So if I ever come off as too direct, please don't take it personal, anybody!

PR #935, opened 2025-04-02
Feature: Mainly for updates of Personalization

ChenYiLins:
### Describe

Refer to TimePage designed by @Jay-o-Way. The function is basically consistent with the original Page.

#### Take an overall view

Take an overall view

#### Picture

Picture

#### Picture mm

Picturemm

#### Solid color

Solid color


Jay-o-Way:
Hm, I have an idea now. I think **Mode** and the settings immediately below that, could also be in one **Expander**. But low priority.

Jay-o-Way:
Honestly, I don't think the ComboButton is appropriate here. User can set things for _both_ themes, it's not one-or-the-other.

Armin2208:
> Honestly, I don't think the ComboButton is appropriate here. User can set things for _both_ themes, it's not one-or-the-other.

Maybe you can realize this with a tab interface in this case.

AutoDarkModeApp/ViewModels/WallpaperPickerViewModel.cs
```diff
@@ -36,21 +84,153 @@ public WallpaperPickerViewModel(IErrorService errorService)
...
+        {
+            SelectMonitor = Monitors[0];
+        }
+
+        if (SelectIndexWallpaperMode == 0)
+        {
+            switch (_builder.Config.WallpaperSwitch.Component.TypeLight)
+            {
+                case WallpaperType.Global:
+                    IsPictureType = true;
```
Jay-o-Way, line 155: What do these variables do? Why do they exist, when they _seem_ to do the same as the previous variable?

AutoDarkModeApp/ViewModels/WallpaperPickerViewModel.cs
```diff
@@ -36,21 +84,153 @@ public WallpaperPickerViewModel(IErrorService errorService)
...
         }
     }
 
+    private void ResetAllOptions()
+    {
+        IsPictureType = false;
+        IsPictureMMType = false;
+        IsSolidColorType = false;
+        IsSpotlightType = false;
+        ImageSettingsCardVisibility = Visibility.Collapsed;
```
Jay-o-Way, line 228: Looks like you created a variable, but why not set `.Visibility` directly? Or use the Enum in the x:Bind?

AutoDarkModeApp/ViewModels/WallpaperPickerViewModel.cs
```diff
@@ -76,4 +256,147 @@ partial void OnIsWallpaperSwitchEnabledChanged(bool value)
...
     }
+
+    partial void OnSelectIndexWallpaperModeChanged(int value)
+    {
+        LoadSettings();
+    }
+
+    partial void OnIsPictureTypeChanged(bool value)
+    {
+        if (SelectIndexWallpaperMode == 0 && value == true)
```
Jay-o-Way, line 267: Is it an idea to use an Enum, instead of zero, one..?

AutoDarkModeApp/ViewModels/WallpaperPickerViewModel.cs
```diff
@@ -76,4 +256,147 @@ partial void OnIsWallpaperSwitchEnabledChanged(bool value)
...
+        {
+            _builder.Save();
+        }
+        catch (Exception ex)
+        {
+            _errorService.ShowErrorMessage(ex, App.MainWindow.Content.XamlRoot, "WallpaperPickPage");
+        }
+    }
+
+    partial void OnIsSpotlightTypeChanged(bool value)
```
Jay-o-Way, line 325: These are very repetitive. There must be a simpler way.

ChenYiLins:
> Honestly, I don't think the ComboButton is appropriate here. User can set things for _both_ themes, it's not one-or-the-other.

Actually, only one theme setting will take effect at a time. In the original code, the option of this ComboBox initializes with the current theme of the computer, which is very intuitive.

But your point of view is actually very correct. In the original Pick an accent color Page, it used the method you mentioned.

If we use ComboBox, it may be ambiguous to users, because users will not know for the first time that different types can be adopted under different themes.

image

Jay-o-Way (commented):
Quickly from my phone

Jay-o-Way:
> Maybe you can realize this with a tab interface in this case.

Or maybe a [Selector bar](https://learn.microsoft.com/en-us/windows/apps/design/controls/selector-bar) ? Let's give this UX a good night sleep.

AutoDarkModeApp/Views/WallpaperPickerPage.xaml
```diff
@@ -9,54 +9,160 @@
...
+                                    BorderThickness="0,1,0,0" />
+                                <TextBlock
+                                    Grid.Row="1"
+                                    VerticalAlignment="Center"
+                                    Style="{StaticResource BodyStrongTextBlockStyle}"
+                                    Text="{helpers:ResourceString Name=WallpaperColor}" />
+                                <Button
+                                    Grid.Row="1"
+                                    MinWidth="120"
+                                    HorizontalAlignment="Right"
```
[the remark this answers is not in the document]
ChenYiLins: Sad  ( ; ´д｀)ゞ This is did happen.

image

AutoDarkModeApp/ViewModels/WallpaperPickerViewModel.cs
ChenYiLins: You are right. It's time to replace it to Enum.

Spiritreader:
The new wallpaper page looks nice! 
Do you know if it's still possible to provide an image preview like with the old UI?

A couple of notes:

> In the original code, the option of this ComboBox initializes with the current theme of the computer, which is very intuitive.

This is really important and mandatory. Otherwise it becomes really confusing when trying to configure it (believe me, I experienced this a lot of times before we implemented this)

ChenYiLins:
## Code quality: Improve WallpaperPickerPage

### Describe

- Change bool variable to enum.
- Add picture preview.
- Add Grid column definition to avoid overlapping controls.
- Increase Border display color.
- On the initialization page, set ThemeMode to follow the current system theme.
- Add converter: EnumToIndexConverter, FlagsToVisibilityConverter.

> This is really important and mandatory. Otherwise it becomes really confusing when trying to configure it (believe me, I experienced this a lot of times before we implemented this)

It's right. We use ComboBox is the better choice. Because in the configuration file, not all settings correspond to Light or Dark Mode.

```yaml
  Component:
    Position: Fill
    Monitors:
    - Id: '\\?\DISPLAY#SKG2725#7&e9a60b&0&UID256#{e6f07b5f-ee97-4a90-b076-33f57bf4eaa7}'
```

### ScreenShots

#### Picture preview

image

#### Avoid overlapping controls

image

#### Colorful border

image

ChenYiLins:
## Code quality: Improve performance

### Describe

- Fix missing enum.
- Prevent the page from being loaded twice because of the change of the value.
- Optimized the code for loading the monitors.
- The method of OnNavigatedFrom is added to View, which makes it possible to remove HandleConfigUpdate when the page leaves. (It's very import! This will cause very serious performance problems! ) (I feel more and more wrong, because it takes a long time to load the page, so I found this problem. (っ °Д °;)っ)

### Supplement

In fact, loading monitors in ViewModel will also cause some performance consumption. This is the test result:
```
Load monitors
Time:104
Time:61
Time:58
Time:55
Time:79

Do not load monitors
Time:22
Time:7
Time:3
Time:26
Time:3
```

Maybe we should separate this method and avoid using x:bind, but GlobalWallpaperPath depends on SelectMonitor (MonitorSettings), which in turn depends on Monitors. My mood: （；´д｀）ゞ

AutoDarkModeApp/Views/WallpaperPickerPage.xaml
```diff
@@ -7,56 +7,209 @@
...
+                                    <RowDefinition Height="Auto" />
+                                    <RowDefinition Height="Auto" />
+                                    <RowDefinition Height="Auto" />
+                                </Grid.RowDefinitions>
+                                <Grid.ColumnDefinitions>
+                                    <ColumnDefinition Width="Auto" />
+                                    <ColumnDefinition Width="*" />
+                                </Grid.ColumnDefinitions>
+
+                                <Border
```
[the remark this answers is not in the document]
ChenYiLins: Because not every block needs a dividing line, such as Spotlight.

AutoDarkModeApp/ViewModels/WallpaperPickerViewModel.cs
```diff
@@ -1,54 +1,216 @@
...
+            WinRT.Interop.InitializeWithWindow.Initialize(openPicker, hWnd);
+            var file = await openPicker.PickSingleFileAsync();
+            if (file != null)
+            {
+                GlobalWallpaperPath = file.Path;
+            }
+        });
     }
 
+    public void OnNaOnNavigatedFrom(NavigationEventArgs e) => StateUpdateHandler.OnConfigUpdate -= HandleConfigUpdate;
```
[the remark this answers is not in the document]
ChenYiLins: At the same time, there is another interesting point here, about INavigationAware? If you use it, I don't know why, when you use the back button of the navigation bar, it doesn't trigger OnNavigatedFrom.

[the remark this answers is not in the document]
Jay-o-Way: The buttons (Pane and Back) in the Navigation control are hidden. Their input is taken over by the two buttons in the XAML TitleBar. Both have a function to connect to the Navigation.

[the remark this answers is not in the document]
ChenYiLins: You are right, that's why we can't use INavigaionAware, I will delete it. And more, I renamed it now.

AutoDarkModeApp/Views/WallpaperPickerPage.xaml
```diff
@@ -7,56 +7,209 @@
...
+
+                <!--  Wallpaper Type  -->
+                <controls:SettingsCard Visibility="{x:Bind ViewModel.IsWallpaperSwitchEnabled, Mode=OneWay}">
+                    <controls:SettingsCard.Description>
+                        <StackPanel Spacing="12">
+                            <RadioButtons SelectedIndex="{x:Bind ViewModel.CurrentDisplayMode, Converter={StaticResource EnumToIndexConverter}, Mode=TwoWay}">
+                                <RadioButton Content="{helpers:ResourceString Name=WallpaperComboBoxItemPicture}" />
+                                <RadioButton Content="{helpers:ResourceString Name=WallpaperComboBoxItemPictureMM}" />
+                                <RadioButton Content="{helpers:ResourceString Name=WallpaperComboBoxItemSolidColor}" />
+                                <RadioButton Content="{helpers:ResourceString Name=ThemePickerTheme11Spotlight}" />
```
Jay-o-Way, line 41: The full string should be "Windows Spotlight". Extra idea: add a Hyperlink to `ms-settings:personalization-background` ?

ChenYiLins: Thanks for your idea. Although not all dividing lines can be merged, because multi-monitors and wallpaper filling methods are only suitable for multi-monitors mode. In the test, I can't seem to use `NavigateUri`, which can't trigger correctly. So I ues `Click`.

ChenYiLins:
## Code quality: Add WindowsSpotlight hyperlink

### Describe

- Merage dividing lines.
- Add a HyperlinkButton to navigate to Windows settings.

### Screenshots

image

AutoDarkModeApp/Helpers/EnumToIndexConverte.cs
Jay-o-Way, line 1: Filename: missing an "R"

AutoDarkModeApp/Views/WallpaperPickerPage.xaml
ChenYiLins: I found [this](https://github.com/microsoft/microsoft-ui-xaml/issues/10428).

AutoDarkModeApp/Helpers/EnumToIndexConverte.cs
ChenYiLins: OK! Fixed.

AutoDarkModeApp/Views/WallpaperPickerPage.xaml
```diff
@@ -142,7 +146,7 @@
                             <Grid RowSpacing="12" Visibility="{x:Bind ViewModel.CurrentDisplayFlags, Converter={StaticResource FlagsToVisibilityConverter}, ConverterParameter=ShowColorSettings, Mode=OneWay}">
                                 <Grid.ColumnDefinitions>
                                     <ColumnDefinition Width="*" />
-                                    <ColumnDefinition Width="32" />
+                                    <ColumnDefinition Width="36" />
```
Jay-o-Way, line 149: Why not use **Auto**?

AutoDarkModeApp/Views/WallpaperPickerPage.xaml
```diff
@@ -18,12 +19,15 @@
...
-                <controls:SettingsCard Header="{helpers:ResourceString Name=WallpaperToggleSwitch}">
+                <controls:SettingsCard Header="{helpers:ResourceString Name=WallpaperToggleSwitch}" HeaderIcon="{ui:FontIcon Glyph=&#xE91B;}">
                     <ToggleSwitch IsOn="{x:Bind ViewModel.IsWallpaperSwitchEnabled, Mode=TwoWay}" />
                 </controls:SettingsCard>
 
                 <!--  Wallpaper Switch Theme  -->
-                <controls:SettingsCard Header="{helpers:ResourceString Name=WallpaperComboBoxMode}" Visibility="{x:Bind ViewModel.IsWallpaperSwitchEnabled, Mode=OneWay}">
+                <controls:SettingsCard
+                    Header="{helpers:ResourceString Name=WallpaperComboBoxMode}"
+                    HeaderIcon="{ui:FontIcon Glyph=&#xE790;}"
```
Jay-o-Way, line 29: How do you feel about icon E793? I think this fits for the dark/light context.

ChenYiLins: To be honest, I referred to the official applications of Win11, such as calculator and notepad.

image

And I try to ues this icon in other place.

image

Jay-o-Way: Good reason!

ChenYiLins:
## Code quality: Renameed

### Describe

- Buttons should convey their meaning through more detailed words, rather than icons.
- Rename some names to make them more uniform.

### Screenshots

image

ChenYiLins:
## Feature: Mainly for updates of ColorizationPage

### Describe

Add ColorizationPage.

### Screenshots

image

gif

ChenYiLins:
> Idea: if it were me, I would say [ Color ... Select color ]

@Jay-o-Way I found a example.

image

Spiritreader:
> ## Code quality: Improve performance
> ### Describe
> 
>     * Fix missing enum.
> 
>     * Prevent the page from being loaded twice because of the change of the value.
> 
>     * Optimized the code for loading the monitors.
> 
>     * The method of OnNavigatedFrom is added to View, which makes it possible to remove HandleConfigUpdate when the page leaves. (It's very import! This will cause very serious performance problems! ) (I feel more and more wrong, because it takes a long time to load the page, so I found this problem. (っ °Д °;)っ)
> 
> 
> ### Supplement
> 
> In fact, loading monitors in ViewModel will also cause some performance consumption. This is the test result:
> 
> ```
> Load monitors
> Time:104
> Time:61
> Time:58
> Time:55
> Time:79
> 
> Do not load monitors
> Time:22
> Time:7
> Time:3
> Time:26
> Time:3
> ```
> 
> Maybe we should separate this method and avoid using x:bind, but GlobalWallpaperPath depends on SelectMonitor (MonitorSettings), which in turn depends on Monitors. My mood: （；´д｀）ゞ

Yeah, getting monitor information from the windows API is unexplicably slow for some reason, which is why there is this caching mechanism. Previously, loading the monitor settings was done asynchronously on page open. Then the page opened quickly  and the monitor strings were loaded in deferred.
Is that no longer an option with WinUi3?

Afaik there's no way to speed this operation up, so at some point getting the name string is necessary.

ChenYiLins:
> Yeah, getting monitor information from the windows API is unexplicably slow for some reason, which is why there is this [caching mechanism](https://github.com/ChenYiLins/Windows-Auto-Night-Mode/blob/79e51e4295c6b74bff2461317ee3c4887ac35895/AutoDarkModeLib/ComponentSettings/Base/WallpaperSwitchSettings.cs#L76). Previously, loading the monitor settings was done asynchronously on page open. Then the page opened quickly and the monitor strings were loaded in deferred. Is that no longer an option with WinUi3?
> 
> Afaik there's no way to speed this operation up, so at some point getting the name string is necessary.

Actually, maybe I'm a little picky. IIn WinUI, we can completely rely on loading asynchronously, but we need to add something to View. Fortunately, the current processing method will not cause too much performance burden.

tooomm:
> Idea: if it were me, I would say [ Color ... Select color ]

image

I also feel the wording of the description and the button is not ideal.
`Accent color                  [ ] [ Select color ]` or `[ Color picker ]`

Spiritreader:
https://github.com/user-attachments/assets/693a35d1-d5f1-4693-83e8-07436c90f31e

Changing the monitor selection does not update the wallpaper path or preview

ChenYiLins:
## Feature: Add the missing function

### Describe

- Switching monitors options will now match the correct wallpaper.
- The unconnected monitors can be removed normally.
- Changed the method of loading the monitors. Due to asynchronous execution, there is almost no additional performance overhead for loading pages.
- Changed the display mode of TeachingTip, and now it will automatically close if it loses focus.

### Screenshots

image

image

Spiritreader (approved):

PR #940, opened 2025-04-13
Update GitHub links to new URL after org creation

tooomm:
Hope I spotted all places :)


Jay-o-Way:
Hey @tooomm thanks for this! Actually, I see a few things. First, we are currently in the progress of porting to WinUI 3. Since you based this PR on the main branch, it would be better to target the `winui` branch. However, I see a bigger picture: so many of the same code in different places. This is an issue that should be solved on a larger scope. Using constants in a central/global place, and using more central functions/methods...

###### disclaimer: I'm not an owner, just helping out since lately

tooomm:
Recreated the PR with the correct target branch: #943

Having such values in constants sounds great!

PR #950, opened 2025-04-15
Winui3-autostart-fix

### LINKED ISSUE #946: Regression: Autostart Handling
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/8daffc4006f3d2ed2eb000207adaacc5d165dd35/AutoDarkModeApp/App.xaml.cs

This block here in the prevous `App.xaml.cs` is also very important

```c#
            //only run at first startup
            if (Settings.Default.FirstRun)
            {
                //enable autostart
                AutostartHandler.EnableAutoStart(null);

               .
               .
               .
            }
            else
            {
                // validate autostart always if firstrun has completed
                AutostartHandler.EnsureAutostart(null);
            }
```

https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/master/AutoDarkModeApp/Handlers/AutostartHandler.cs

it ensures that ADM is correctly added to autostart when the app is run. Every subsequent app start after the initial run should call the `EnsureAutostart` method of the AutostartHandler to ensure that user tampering does disable autostart.
This is really important, as user tampering happens a lot more than you would expect.

Jay-o-Way:
Fixes #946
Created with Copilot!

### DIFF
--- a/AutoDarkModeApp/App.xaml.cs
+++ b/AutoDarkModeApp/App.xaml.cs
@@ -100,6 +100,24 @@ public partial class App : Application
 
         await SetApplicationLanguageAsync();
 
+        // Autostart logic
+        var localSettings = App.GetService<ILocalSettingsService>();
+        var isFirstRun = await localSettings.ReadSettingAsync<bool>("FirstRun");
+
+        if (isFirstRun)
+        {
+            // Enable autostart on first run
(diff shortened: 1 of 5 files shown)


Jay-o-Way:
Not as simple as copy-paste. I don't know how Handlers work. Sorry.

PR #947, opened 2025-04-15
Feature: Mainly for updates of ThemePickerPage

ChenYiLins:
### Describe

Add ThemePickerPage.

### Screenshots

image

image

#### Theme

image

#### Pick theme

image

#### Ignore

image

PR #954, opened 2025-04-23
Code quality: Streamlined navigation and more improvements

ChenYiLins:
### Describe

- Streamlined NavigationService and replaced ItemInvoked with a more concise SelectionChanged.
- Delete ViewModel for MainWindow, put NavigationView logic into NavigationService.
- Remove NavigationHelper, use tag in MainWindow.
- Change Apps to System Areas.

### Supplement

The reason for choosing to keep PageService is that it provides the mapping from ViewModel to Page, where page types can be managed centrally and pages can be added/removed dynamically (if needed later).


AutoDarkModeApp/App.xaml.cs
```diff
@@ -100,8 +97,8 @@ protected async override void OnLaunched(LaunchActivatedEventArgs args)
 
         await SetApplicationLanguageAsync();
 
-        var mainViewModel = App.GetService<MainViewModel>();
-        MainWindow = new MainWindow(mainViewModel);
+        var navigationService = App.GetService<INavigationService>();
```
Jay-o-Way, line 100: Sorry, I'm not so familiar with services, but for what is this still needed?

AutoDarkModeApp/MainWindow.xaml
```diff
@@ -51,39 +48,40 @@
               TODO: Update item icons by updating FontIcon.Glyph properties.
               https://docs.microsoft.com/windows/apps/design/style/segoe-fluent-icons-font#icon-list
                 -->
-                <NavigationViewItem helpers:NavigationHelper.NavigateTo="AutoDarkModeApp.ViewModels.TimeViewModel" Content="{helpers:ResourceString Name=NavbarTime}">
+                <NavigationViewItem Content="{helpers:ResourceString Name=NavbarTime}" Tag="AutoDarkModeApp.ViewModels.TimeViewModel">
```
Jay-o-Way, line 51: Interesting, I thought a **Tag** was a simple string. Can you tell/show how this works?

AutoDarkModeApp/MainWindow.xaml.cs
```diff
@@ -29,8 +28,8 @@ public MainWindow(MainViewModel viewModel)
 
         Title = Debugger.IsAttached ? "Auto Dark Mode Dev" : "Auto Dark Mode";
 
-        ViewModel.NavigationService.Frame = NavigationFrame;
-        ViewModel.NavigationService.InitializeNavigationView(NavigationViewControl);
+        _navigationService.Frame = NavigationFrame;
```
Jay-o-Way, line 31: Same thing - for what is this still needed?

AutoDarkModeApp/MainWindow.xaml
```diff
@@ -11,15 +11,14 @@
     <Window.SystemBackdrop>
         <MicaBackdrop />
     </Window.SystemBackdrop>
-    <Grid x:Name="RootGrid">
+    <Grid x:Name="RootGrid" PointerPressed="RootGrid_PointerPressed">
```
Jay-o-Way, line 14: Sorry, but this is not working. It updates the **Frame**, but doesn't touch the **Navigation** control at all.

AutoDarkModeApp/Services/NavigationService.cs
```diff
@@ -49,105 +27,52 @@ public NavigationService(IPageService pageService)
     public void InitializeNavigationView(NavigationView navigationView)
     {
         _navigationView = navigationView;
-        _navigationView.ItemInvoked += OnItemInvoked;
-    }
+        _navigationView.SelectionChanged += OnSelectionChanged;
```
Jay-o-Way, line 30: In the examples I've seen, this property is usually set in the XAML side. But both are valid.

Jay-o-Way (changes requested):
Thanks for starting on this! Suggestion: If you've started working on an issue, give us a hint, so we don't do the same issue at the same time, but independently :)

As for the PR: few comments :)

Jay-o-Way:
Probably out of this PR, but when I hover the main window, I see a tooltip that says "Alt + Left" (in my language). This must be a tooltip for a GoBack hotkey, but it shouldn't be shown. Do you know the origin of this?

AutoDarkModeApp/MainWindow.xaml.cs
```diff
@@ -29,8 +28,8 @@ public MainWindow(MainViewModel viewModel)
...
+        _navigationService.InitializeNavigationView(NavigationViewControl);
 
         RootGrid.KeyboardAccelerators.Add(BuildKeyboardAccelerator(VirtualKey.Left, VirtualKeyModifiers.Menu));
```
Jay-o-Way, line 34: (Out of scope for PR, but) This places a ToolTip across the whole window. Not ideal.

AutoDarkModeApp/MainWindow.xaml
ChenYiLins: I guess you saw the code in WinUIGallery. In fact, it just didn't handle namespace directly in xaml, and we can do the same.
```c#
    private void NavigationView_SelectionChanged(Microsoft.UI.Xaml.Controls.NavigationView sender, Microsoft.UI.Xaml.Controls.NavigationViewSelectionChangedEventArgs args)
    {
        if (args.IsSettingsSelected)
        {
            contentFrame.Navigate(typeof(SampleSettingsPage));
        }
        else
        {
            var selectedItem = (Microsoft.UI.Xaml.Controls.NavigationViewItem)args.SelectedItem;
            if (selectedItem != null)
            {
                string selectedItemTag = ((string)selectedItem.Tag);
                // Look here
                sender.Header = "Sample Page " + selectedItemTag.Substring(selectedItemTag.Length - 1);
                string pageName = "WinUIGallery.SamplePages." + selectedItemTag;
                Type pageType = Type.GetType(pageName);
                contentFrame.Navigate(pageType);
            }
        }
    }
```

Jay-o-Way: Yes, I look at https://learn.microsoft.com/windows/apps/design/controls/navigationview.

> it just didn't handle namespace directly in xaml

Maybe because the following is not using a TypeOf?

```cs
        else if (args.SelectedItemContainer?.Tag is string pageKey)
        {
            NavigateTo(pageKey);
```

ChenYiLins: > Maybe because the following is not using a TypeOf?
> 
> ```cs
>         else if (args.SelectedItemContainer?.Tag is string pageKey)
>         {
>             NavigateTo(pageKey);
> ```

No, PageKey has been transformed, which is really just a matter of namespace.

```c#
    public bool NavigateTo(string pageKey, object? parameter = null, bool clearNavigation = false)
    {
        if (_frame == null)
            return false;
        // Here
        var pageType = _pageService.GetPageType(pageKey);

        if (clearNavigation)
        {
            _frame.BackStack.Clear();
        }

        return _frame.Navigate(pageType, parameter);
    }
```
As for WinUIGallery itself, it uses a lot of selection structures and passes tags as parameters.

```c#
            var selectedItem = args.SelectedItemContainer;
            if (selectedItem == AllControlsItem)
            {
                if (rootFrame.CurrentSourcePageType != typeof(AllControlsPage))
                {
                    Navigate(typeof(AllControlsPage));
                }
            }
            else if (selectedItem == Home)
            {
                if (rootFrame.CurrentSourcePageType != typeof(HomePage))
                {
                    Navigate(typeof(HomePage));
                }
            }
            else if (selectedItem == ColorItem)
            {
                Navigate(typeof(ItemPage), "Color");
            }
```

AutoDarkModeApp/App.xaml.cs
ChenYiLins: NavigationService is used to deal with the actual navigation logic, which well separates the navigation logic from the MainWindow, otherwise we will be closely linked with the MainWindow when dealing with the navigation logic in other places.

AutoDarkModeApp/MainWindow.xaml.cs
ChenYiLins: Since we have chosen the separation of navigation logic, we should not directly let NavigationService access the MainWindow and indirectly pass the Frame to the service.

AutoDarkModeApp/MainWindow.xaml
ChenYiLins: In fact, it is really working hard. ≡(▔﹏▔)≡ It is only used to ensure that the mouse's side keys can work normally (go back and forward), but thanks to your reminder, I found that the Header doesn't seem to be updated normally after we discarded ViewModel.

AutoDarkModeApp/Services/NavigationService.cs
ChenYiLins: Yep, but this is also to separate the navigation logic.

AutoDarkModeApp/MainWindow.xaml.cs
ChenYiLins: Uh...（；´д｀）ゞ I found out that this doesn't seem to be our expected behavior.

AutoDarkModeApp/MainWindow.xaml
Jay-o-Way: Don't forget the Selected Item too.

AutoDarkModeApp/MainWindow.xaml.cs
Jay-o-Way: > Since we have chosen the separation of navigation logic, we should not directly let NavigationService access the MainWindow and indirectly pass the Frame to the service.

I think I missed the moment this decision was made.

AutoDarkModeApp/Services/ActivationService.cs
ChenYiLins, line 1: Good change! Do we need to use Primary Constructors in more Service?

Jay-o-Way: I'm only applying a few suggestions I see, nothing more 🤣

AutoDarkModeApp/MainWindow.xaml.cs
Jay-o-Way: 1. Let's check if the function of the mouse buttons is not already built-in to WinUI 3 💡🤓
2. If not - found this page: https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.navigationview?view=windows-app-sdk-1.7#examples
We can see how that works? Note that we should find the button(s) in the XAML **TitleBar** control, because the buttons in the NavigationView are hidden.

ChenYiLins: Do you mean AccessKey? But what we need is Hotkey.

I went to watch the code in WinUIGallery, which uses a more complicated processing method. For details, please see [App.xaml.cs](https://github.com/microsoft/WinUI-Gallery/blob/2d50615b71e80d153d13b5fb71636af10f84b052/WinUIGallery/App.xaml.cs#L92) and [NavigationHelper](https://github.com/microsoft/WinUI-Gallery/blob/main/WinUIGallery/Helpers/NavigationHelper.cs). I don't think this is an excellent reference object.

I will try to improve the code.

ChenYiLins:
## Code quality: Perfect keyboard and mouse operation

### Describe

- Merge keyboard and mouse operations into the Service, and the MainWindow is now clean.ヾ(≧▽≦*)o
- Now we can use not only Alt + ←, but also Alt + →.
- Fix the stuck problem when the App is closed in debugging mode.
- Hide unexpected ToolTip.

### Screenshots

image

ChenYiLins:
## Code quality: Add subtree of items

### Describe

- Add subtree of items for Personalization item.
- Replace FontIcon to CommunityToolkit's FontIcon.

### Screenshots

image

AutoDarkModeApp/MainWindow.xaml.cs
Jay-o-Way: > I went to watch the code in WinUIGallery

Yeah, WinUI Gallery is nice, but it's also made by people. It's not holy by itself. A quick look at the [blame of the NavigationHelper file](https://github.com/microsoft/WinUI-Gallery/blame/main/WinUIGallery/Helpers/NavigationHelper.cs) tells me that the code/file is 7 (seven!) years old...
Also note the [draft pr](https://github.com/microsoft/WinUI-Gallery/pull/1797) they have, that changes a lot about the MainWindow and related things.

Jay-o-Way:
Interesting choice to add the sub-items. Now I think there are two paths that eventually do the exact same thing. You can click in the Nav, or you can click on the page. It feels redundant to me. Need to know what @Armin2208 and @Spiritreader thing about this.

Jay-o-Way:
### Bug

This happened when I clicked on the **Theme** item in the menu
```
Exception thrown: 'System.ArgumentException' in AutoDarkModeApp.dll
An exception of type 'System.ArgumentException' occurred in AutoDarkModeApp.dll but was not handled in user code
Page not found: AutoDarkModeApp.ViewModels.ThemePickerViewModel. Did you forget to call PageService.Configure?
```
image

ChenYiLins:
> ### Bug
> This happened when I clicked on the **Theme** item in the menu. Note that the **SettingsCard** on the page has no **Click** event linked to it (yet) but #947 will solve that.

Yep! This PR is based on #947, so after the merger of #947, these functions can work normally.

ChenYiLins:
> Interesting choice to add the sub-items. Now there are two paths that do the exact same thing. You can click in the Nav, or you can click on the page. This is redundant. Need to know what @Armin2208 and @Spiritreader think about this.

I made this change to make the navigation structure more intuitive and clear. The reason for keeping personalized items is that users can see the descriptions of different items on this page.

Jay-o-Way:
> > Interesting choice to add the sub-items. Now there are two paths that do the exact same thing. You can click in the Nav, or you can click on the page. This is redundant. Need to know what @Armin2208 and @Spiritreader think about this.
> 
> I made this change to make the navigation structure more intuitive and clear. The reason for keeping personalized items is that users can see the descriptions of different items on this page.

Yes, but - at the moment - each Nav Item is functionally **identical** to the clickable Setting Cards. That's just totally unnatural and unneeded. I suggest to choose one solution.

ChenYiLins:
> Yes, but - at the moment - each Nav Item is functionally **identical** to the clickable Setting Cards. That's just totally unnatural and unneeded. I suggest to choose one solution.

Uh... It's just like the way WinUI Gallery uses it.

Spiritreader:
My general recommendation (and most likely @Armin2208's) is to keep navigation functionality as close to Windows defaults as possible.

In case Windows does something that is unintuitive, Armin mostly has a pretty good radar on what to keep, so I'll get in contact with him to ask for further clarification

AutoDarkModeApp/MainWindow.xaml.cs
Spiritreader: > Actually, I think separation logic is a good thing, right? Let the MainWindow not focus on doing too many things,

I would tend to agree with this. MainWindow became too much of a monolith in the old UI, making it harder to change UI behavior and look&feel later on.

@ChenYiLins do you think that there is added maintenance cost of choosing a separated approach?
My potential concern is that custom navigation logic isn't strictly requried, but from the look of it it can be used to achieve potentially better results.

Also, some testing will need to be done on touch devices, because we've had touch users in the past that had bad experiences due to bugged navigation with the previous UI framework (ModernWPF)

Jay-o-Way: > some testing will need to be done on touch devices

I have a touch screen and it works lovely :)

Armin2208:
Firstly, great work from you guys! The screenshots look promising, and they make me happy, considering we're finally processing to an Auto Dark Mode 11 with a new and modern User Interface.

Also, thanks for having attention to the details. Things like Alt + ← & Alt + → are really improving the user experience in the long term. You are doing a great job overall.

**As a user, I would expect the following keyboard shortcut logic:**

- Alt + Left = going back to the previous page (like a webbrowser)
- Alt + Right = going forward, to the previous page (like a webbrowser)
- Alt + Up = going one page in navigation tree up
- Alt + Down = going one page in navigation tree down.
- Ctrl + Tab = going one page in navigation tree down.
- Strg + Shift + Tab = going one page in navigation tree up.

From the video, I can't really tell what which button does in the current implementation. I also don't know what "XButtonUp" is.

**Now speaking of the subtree menus on the navigation tree:**
They should not be included. Here is why:

- The user have either the choice of using a theme or let Auto Dark Mode manage background / mouse / accent. We need to show the user that he has a choice. The navigation tree removes this sense of choice.
- If a theme is chosen, there is no way to set background / mouse / accent anymore. We gray them out, with a message, that they are disabled due to the use of a Windows theme. This also can't be reflected by the navigation tree.

Due to these two important factors, I recommend not to implement the navigation tree in the navigation bar. Nice idea, but sadly not a good choice from a UX perspective. But still, thanks for trying out new stuff and experimenting.

Btw: I don't know if you already have implemented the red message box in the Page Personalization, that the three entries background / mouse / accent are disabled, due to the use of a Windows theme. Maybe we can extend this red message box with a new button, which does disable Windows theme for the user - because it could be that the user doesn't know how to disable it. A little idea to extend the easiness of usage.

Thanks! Looking forward to seeing more work of you guys. You can always contact me regarding UX thoughts.

ChenYiLins:
> **Now speaking of the subtree menus on the navigation tree:** They should not be included. Here is why:
> 
> * The user have either the choice of using a theme or let Auto Dark Mode manage background / mouse / accent. We need to show the user that he has a choice. The navigation tree removes this sense of choice.
> * If a theme is chosen, there is no way to set background / mouse / accent anymore. We gray them out, with a message, that they are disabled due to the use of a Windows theme. This also can't be reflected by the navigation tree.
> 
> Due to these two important factors, I recommend to not implement the navigation tree in the navigation bar. Nice idea, but sadly not a good choice from a UX perspective. But still, thanks for trying out new stuff and experimenting.

This is a very important point, and the existence of subtree in navigation tree will cause serious ambiguity. 

> Also, there's the question for the personalization page in peraticular.
> If theme mode is enabled, it should no longer be possible to open the non-theme mode pages and vice versa. The nav bar must also reflect this.

> Btw: I don't know if you already have implemented the red message box in the Page Personalization, that the three entries background / mouse / accent are disabled, due to the use of a Windows theme. Maybe we can extend this red message box with a new button, which disables Windows theme for the user - because it could be that the user doesn't know how to disable it. A little idea to extend the easiness of usage.

Ah ah ah ah, thank everyone very much for reminding me. I forgot this from beginning to end.（；´д｀）ゞ

AutoDarkModeApp/MainWindow.xaml.cs
ChenYiLins: > do you think that there is added maintenance cost of choosing a separated approach with your Navigation Helper? My potential concern is that custom navigation logic may not strictly be required and changes in the future to WinUI that affect this code will have to be fixed by us manually. But from the looks of it it can be used to achieve potentially better results with different nav approaches.

In fact, I think the current NavigationService is relatively simple, with no complicated API mixed in and too many functions implemented. 

I guess @Jay-o-Way  has been puzzled by why I am obsessed with using services. Although I have separated viewmodel from MainWindow, in the final analysis, our program framework is based on MVVM. (In order to ensure the consistency of the whole program, we choose to use Service instead of separate Helper. At the same time, because of the existence of PageService, choosing Helper will cause great procedural fragmentation.) MVVM is not reflected in WinUI Gallery, and also includes MS Learn examples. Benefits to service, as a photo:

image

In fact, I don't think I did it right either. Service should be used in viewmodels (Just like PageService maps viewmodels to views, because MVVM pattern supposes that the viewmodels doesn’t know about views). Although the current method is not absolutely correct, I think I have chosen a scheme that balances convenience and complexity. φ(*￣0￣)

At the same time, the navigation logic implemented in NavigationService is actually in strict accordance with MS Learn examples. I believe that there will be no WinUI changes that will have a serious impact on this part in the future.

ChenYiLins:
> Btw: I don't know if you already have implemented the red message box in the Page Personalization, that the three entries background / mouse / accent are disabled, due to the use of a Windows theme. Maybe we can extend this red message box with a new button, which disables Windows theme for the user - because it could be that the user doesn't know how to disable it. A little idea to extend the easiness of usage.

Add it in #947 now.

AutoDarkModeApp/MainWindow.xaml.cs
Spiritreader: Thank you for the detailed write up!
I now understand the design decision a lot better, and agree that having too many helpers in general is clunky.

> In fact, I don't think I did it right either.

I don't think that doing things "strictly correct" is necessarily always the right way, the architecture needs to fit the use case. So I agree on your stance on balancing convenience with complexity

Jay-o-Way:
I like where this is going 🙂 At what point do we want to pull this into the `winui` branch?

ChenYiLins:
## Code quality: Format xaml

### Description

Using blank line separation is more conducive to quickly scanning the file structure. _Symmetry can quickly identify the nesting range of tags, especially in complex XAML, which is easier to maintain._

## Code quality: Operation health

### Description

- Replace bool variable with enum. At the same time, add converter in order to convert enum into Visibility.
- Change [ObservableProperty] to adapt to Aot.
- Changing the error output to ViewModel, the Page text may cause an error prompt that is easily misunderstood.
- Update StateUpdateHandler, now has the function of debounce. At the same time, the events that are cleared when the watcher is started, so it is not necessary to manually clear when leaving the navigation. At the same time, some codes are improved, and now it is more secure and reliable.

ChenYiLins:
> **As a user, I would expect the following keyboard shortcut logic:**
> 
> * Alt + Left = going back to the previous page (like a webbrowser)
> * Alt + Right = going forward, to the previous page (like a webbrowser)
> * Alt + Up = going one page in navigation up
> * Alt + Down = going one page in navigation down.
> * Ctrl + Tab = going one page in navigation down.
> * Ctrl + Shift + Tab = going one page in navigation up.
> 
> From the video, I can't really tell what which button does in the current implementation. I also don't know what "XButtonUp" is.

In fact, WinUI itself provides enough keyboard support, and the “Tab” key plays an important role in it, just like the setting of Win11! We may not need to add too many shortcut keys? At the same time, I forgot to mention that "XButtonUp" and "XButtonDown" are the side keys of the mouse, just like Logitech's GPW.

ChenYiLins:
## Test: Let the time follow the system settings

### Description

Let the time-related components of ADMApp follow the date format of the system.

Jay-o-Way:
> Let the time follow the system settings

Works! Great job! Any change to the system format setting requires a re-load of the app/viewmodel, am I right?
image

Jay-o-Way (approved):
🚀

ChenYiLins:
> Works! Great job! Any change to the system format setting requires a re-load of the app/viewmodel, am I right? !

Yep. I don’t think it’s necessary to constantly check if the system’s time format has changed. Honestly, I doubt any user would be so bored as to tweak the time format every other second ( pretty sure about that ( •̀ ω •́ )✧).

PR #956, opened 2025-04-26
Feature: Service Start Handling and Autostart Handling

ChenYiLins:
### Describe

Restore Service Start Handling. Due to the complexity of Dialog in WinUI, code waiting for XamlRoot was added (the attempt to use Actived failed).

#### Different States

image
image

#### When everything is working properly

image

#### Waiting for service start

image

#### When the service does not respond

image

#### Start Service after App starts

image

### DIFF
--- a/AutoDarkModeApp/Services/ActivationService.cs
+++ b/AutoDarkModeApp/Services/ActivationService.cs
@@ -82,6 +82,7 @@ public class ActivationService : IActivationService
                     },
                 },
             };
+            return;
         }
         else if (serviceStartIssued && verifyResult)
         {
@@ -173,13 +174,21 @@ public class ActivationService : IActivationService
     private static async Task<bool> VerifyServiceStartupAsync()
     {
         const int maxRetries = 5;
-        string response = null!;
+        ApiResponse response = null!;
         for (int i = 0; i < maxRetries; i++)
         {
-            response = await Task.Run(() => MessageHandler.Client.SendMessageAndGetReply(Command.Alive));
+            response = await Task.Run(() => ApiResponse.FromString(MessageHandler.Client.SendMessageAndGetReply(Command.Alive)));
             await Task.Delay(1000);
         }
-        return !response.Contains(StatusCode.Timeout);
+
+        if (response.StatusCode == StatusCode.Timeout)
+        {
+            return false;
+        }
+        else
+        {
+            return true;
+        }
     }
 
     private async Task SystemTimeFormatAsync()
--- a/AutoDarkModeApp/Services/ActivationService.cs
+++ b/AutoDarkModeApp/Services/ActivationService.cs
@@ -128,7 +128,7 @@ public class ActivationService : IActivationService
 
     private static bool StartService()
     {
-        if (Debugger.IsAttached)
+        if (!Debugger.IsAttached)
         {
             using Mutex serviceRunning = new(false, "330f929b-ac7a-4791-9958-f8b9268ca35d");
             if (serviceRunning.WaitOne(TimeSpan.FromMilliseconds(100), false))
--- a/AutoDarkModeApp/Services/ActivationService.cs
+++ b/AutoDarkModeApp/Services/ActivationService.cs
@@ -1,5 +1,12 @@
-﻿using AutoDarkModeApp.Contracts.Services;
+﻿using System.Diagnostics;
+using AutoDarkModeApp.Contracts.Services;
+using AutoDarkModeApp.Helpers;
+using AutoDarkModeApp.Utils.Handlers;
 using AutoDarkModeApp.ViewModels;
+using AutoDarkModeLib;
+using AutoDarkModeSvc.Communication;
+using Microsoft.UI.Dispatching;
+using Microsoft.UI.Xaml.Controls;
 
 namespace AutoDarkModeApp.Services;
 
@@ -24,6 +31,60 @@ public class ActivationService : IActivationService
 
         // Activate the MainWindow.
         App.MainWindow.Activate();
+
+        // Start the service and handle UI flow
+        var serviceStartIssued = StartService();
+        ContentDialog loadingDialog = null!;
+
+        if (serviceStartIssued)
+        {
+            await WaitForXamlRootAsync();
+
+            loadingDialog = new ContentDialog
+            {
+                Title = "StartupLaunchingServiceTitle".GetLocalized(),
+                Content = new StackPanel
+                {
+                    Orientation = Orientation.Vertical,
+                    Spacing = 32,
+                    Children =
+                    {
+                        new TextBlock { Text = "StartupLaunchingServiceText".GetLocalized() },
+                        new ProgressBar { IsIndeterminate = true },
+                    },
+                },
+                XamlRoot = App.MainWindow.Content.XamlRoot,
+            };
+
+            DispatcherQueue.GetForCurrentThread().TryEnqueue(async () => await loadingDialog.ShowAsync());
+        }
+
+        // Wait for service to become responsive
+        var verifyResult = await VerifyServiceStartupAsync();
+
+        if (serviceStartIssued && !verifyResult)
+        {
+            loadingDialog.Content = new StackPanel
+            {
+                Orientation = Orientation.Vertical,
+                Spacing = 32,
+                Children =
+                {
+                    new InfoBar
+                    {
+                        Title = "errorOcurredTitle".GetLocalized(),
+                        Severity = InfoBarSeverity.Error,
+                        IsOpen = true,
+                        IsClosable = false,
+                        Message = "StartupServiceUnresponsive".GetLocalized(),
+                    },
+                },
+            };
+        }
+        else if (serviceStartIssued && verifyResult)
+        {
+            loadingDialog.Hide();
+        }
     }
 
     private async Task MoveWindowAsync()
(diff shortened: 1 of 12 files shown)


Jay-o-Way:
> Due to the complexity of Dialog

- Can you explain more about this?

ChenYiLins:
>  Can you explain more about this?

To put it simply:

[Link:](https://learn.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/dialogs)
_By default, content dialogs display modally relative to the root [ApplicationView](https://learn.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.applicationview). When you use ContentDialog inside of either an [AppWindow](https://learn.microsoft.com/en-us/uwp/api/windows.ui.windowmanagement.appwindow) or a [XAML Island](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/xaml-islands/xaml-islands), you need to manually set the [XamlRoot](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.uielement.xamlroot) on the dialog to the root of the XAML host._

_The XamlRoot property only gets set when the element in question gets loaded into a visual tree._

Dialog is used in a different way from WPF. If we want to display Dialog in ActivationService, we must wait for XamlRoot to load, otherwise we will receive the error: **This element is already associated with a XamlRoot, it cannot be associated with a different one until it is removed from the previous XamlRoot.**

ChenYiLins:
By the way, this original code can't be used normally when it is migrated intact. HeartBeatOK always has extra news. So I have to choose to use Contains to judge. @Spiritreader Do you know why?

#### Previous code

```c#
        private void WaitForServiceStart()
        {
            string heartBeatOK = MessageHandler.Client.SendMessageWithRetries(Command.Alive, retries: 5);
            if (heartBeatOK == StatusCode.Timeout)
            {
                string error = AdmProperties.Resources.StartupServiceUnresponsive;
                Dispatcher.Invoke(() =>
                {
                    MsgBox msgErr = new(error, AdmProperties.Resources.errorOcurredTitle, "error", "close")
                    {
                    };
                    _ = msgErr.ShowDialog();
                });
                Environment.Exit(-2);
            }
        }
```

#### Debug message

```
Timeout
AdmApiDataRow=The service did not acknowledge the req in time
AdmApiDataRow=System.TimeoutException The operation has timed out.
```

#### New code

```c#
    private static async Task<bool> VerifyServiceStartupAsync()
    {
        const int maxRetries = 5;
        string response = null!;
        for (int i = 0; i < maxRetries; i++)
        {
            response = await Task.Run(() => MessageHandler.Client.SendMessageAndGetReply(Command.Alive));
            await Task.Delay(1000);
        }
        return !response.Contains(StatusCode.Timeout);
    }
```

Jay-o-Way:
To be honest, I don't know how to properly review this. Don't know much about services, or to trigger the UI.
Looking at the MS docs; I do wonder if there is an even better way to show these things, but I don't have the answer to that right now. Take the situation when startup fails, for example. What does the user do then? There are no buttons in the Dialog and the window is blocked?

ChenYiLins:
> To be honest, I don't know how to properly review this. Don't know much about services, or to trigger the UI. Looking at the MS docs; I do wonder if there is an even better way to show these things, but I don't have the answer to that right now. Take the situation when startup fails, for example. What does the user do then? There are no buttons in the Dialog and the window is blocked?

No. In the original method, the program exited with error code -2. But now nothing will happen except this error dialog box. Users can choose to close the window by themselves, and the window is not completely inoperable.

ChenYiLins:
## Feature: Autostart Handling

### Describe

Solve #946.

### Screenshots

image

Jay-o-Way:
Okay, so the window **looks** blocked. With no buttons in the dialog, I still think this can leave users "hanging" confused for a while.

Jay-o-Way:
Thought on a larger scope: is the _service_ absolutely necessary for the _settings_ window to function? If the answer is no, then we need to think about other solutions.

ChenYiLins:
> Okay, so the window **looks** blocked. With no buttons in the dialog, I still think this can leave users "hanging" confused for a while. There could be a bit more info to help users. "Try again" sounds simple, but also a bit vague... How about some buttons like "retry " and "close"? (for example

That makes sense! From the user's point of view, it does not convey a correct message visually. At the same time, it is not difficult to add a button, but is it necessary to let the user wait for a long time? (I think it is really long, and the progress bar is only a visual indication that the task is in progress, and the user has no idea where it is going and how long it will end? )

> Also: check accessibility. Is this keyboard friendly?

The user can close the window by Alt + F4. But I found a very interesting point (っ °Д °;)っ , perhaps a loophole in WinUI. Although we can't do anything with the mouse when the dialog is showed (For other elements in the window), we can do it with the keyboard.

Spiritreader:
> Thought on a larger scope: is the _service_ absolutely necessary for the _settings_ window to function? If the answer is no, then we need to think about other solutions.

The answer to that is yes.
Settings window depends on the service.

ChenYiLins:
> Thought on a larger scope: is the _service_ absolutely necessary for the _settings_ window to function? If the answer is no, then we need to think about other solutions.

Of course, as far as basic services are concerned, although not all setting functions depend on it (such as modifying configuration files), it is the basis of ADM. However, most of the setup functions still need the service to run.

Spiritreader:
> By the way, this original code can't be used normally when it is migrated intact. HeartBeatOK always has extra news. So I have to choose to use Contains to judge. @Spiritreader Do you know why?
> 
> #### Previous code
> ```cs
>         private void WaitForServiceStart()
>         {
>             string heartBeatOK = MessageHandler.Client.SendMessageWithRetries(Command.Alive, retries: 5);
>             if (heartBeatOK == StatusCode.Timeout)
>             {
>                 string error = AdmProperties.Resources.StartupServiceUnresponsive;
>                 Dispatcher.Invoke(() =>
>                 {
>                     MsgBox msgErr = new(error, AdmProperties.Resources.errorOcurredTitle, "error", "close")
>                     {
>                     };
>                     _ = msgErr.ShowDialog();
>                 });
>                 Environment.Exit(-2);
>             }
>         }
> ```
> 
> #### Debug message
> ```
> Timeout
> AdmApiDataRow=The service did not acknowledge the req in time
> AdmApiDataRow=System.TimeoutException The operation has timed out.
> ```
> 
> #### New code
> ```cs
>     private static async Task<bool> VerifyServiceStartupAsync()
>     {
>         const int maxRetries = 5;
>         string response = null!;
>         for (int i = 0; i < maxRetries; i++)
>         {
>             response = await Task.Run(() => MessageHandler.Client.SendMessageAndGetReply(Command.Alive));
>             await Task.Delay(1000);
>         }
>         return !response.Contains(StatusCode.Timeout);
>     }
> ```

What do you mean by extra news? Do you mean the extra data rows?
You don't have to necessarily check for contains.

The messaging client returns a serialized `ApiResponse` object which has static methods to unwrap the serialized string.
I don't exactly remember why the old code behaved like that (there may have been changes during API development that were not carried over to the UI

For the new code I would recommend `ApiResponse.FromString` which is part of the symlinked "Communication" namespace via `using AutoDarkModeSvc.Communication;`

https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/8daffc4006f3d2ed2eb000207adaacc5d165dd35/AutoDarkModeSvc/Communication/Command.cs

Armin2208:
Crazy good addition, thanks.

Jay-o-Way:
Is this PR ready?

ChenYiLins:
> Is this PR ready?

As far as this small part is concerned, it seems that it have be completed.

Spiritreader:
will check

AutoDarkModeApp/Services/ActivationService.cs
```diff
@@ -24,6 +33,84 @@ public async Task ActivateAsync(object activationArgs)
...
+
+            await SystemTimeFormatAsync();
+
+            await AddJumpListAsync();
+
+            await _localSettings.SaveSettingAsync("NotFirstRun", true);
+        }
+        else
+        {
+            AutostartHandler.EnableAutoStart(App.MainWindow.Content.XamlRoot);
```
Spiritreader, line 105: On a non-first-run, this needs to be "EnsureAutoStart" instead of "EnableAutoStart"

These two functions differ. The first one instructs the service to create an autostart entry when ADM is launched for the first time.

EnsureAutoStart will only re-enable AutoStart capabilities if it has been erroneously removed by the user (e.g not through the AdmApp)

https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/8daffc4006f3d2ed2eb000207adaacc5d165dd35/AutoDarkModeApp/App.xaml.cs#L141


https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/8daffc4006f3d2ed2eb000207adaacc5d165dd35/AutoDarkModeApp/Handlers/AutostartHandler.cs#L33C9-L48C10
```
public static void EnsureAutostart(Window owner)
        {
            ApiResponse result = new()
            {
                StatusCode = StatusCode.Err,
                Message = "error in frontend: EnsureAutostart()"
            };
            try
            {
                _ = MessageHandler.Client.SendMessageAndGetReply(Command.ValidateAutostart);
            }
            catch (Exception ex)
            {
                ErrorMessageBoxes.ShowErrorMessageFromApi(result, ex, owner);
            }
        }
```

Spiritreader (changes requested):

Spiritreader (approved):
I've made a couple of updates to the branch that is sort of related to service start & autostart handling:

- I fixed the base paths and restored a proper directory structure, such that launching the service application with regards to its physical file location is now consistent
- I restored the mutex behavior of the app, because it must not be able to start twice. Now instead, if launched twice, it will bring the current active window into the foreground
- I revamped the internal structure of ADM into `core` and `ui` and adjusted the helpers such that the execution directories for the respective components are resolved completely.

In addition, I fixed the following:
- The service start launch window was always being shown for 5 seconds, regardless of whether the service actually started or not. I added an early return to the `VerifyServiceStartupAsync()` method which brings startup performance in line with how it previously behaved


I will merge this shortly

ChenYiLins:
> I've made a couple of updates to the branch that is sort of related to service start & autostart handling:

Thank you very much! For the file conflict problem after publish, I have actually tried to solve it in another branch, just add these for svc:

```csproj
    <SelfContained>true</SelfContained>
    <Private>true</Private>
```

Of course, I think it is a good thing to separate the components, unless it really increases the size of the program to a great extent. φ(*￣0￣)

Spiritreader:
> > I've made a couple of updates to the branch that is sort of related to service start & autostart handling:
> 
> Thank you very much! For the file conflict problem after publish, I have actually tried to solve it in another branch, just add these for svc:
> 
> ```
>     <SelfContained>true</SelfContained>
>     <Private>true</Private>
> ```
> 
> Of course, I think it is a good thing to separate the components, unless it really increases the size of the program to a great extent. φ(*￣0￣)

Yeah, every self contained setting will cause size inflation.

I would like to keep the publishing options as it because it works with the update system. 

My idea is to separate the UI from the rest of ADM because it makes the most sense in terms of separation of responsibilities.
We could make the app self contained as there are no shared assemblies, which would reduce the number of files in that directory

PR #958, opened 2025-04-30
Code quality: Operation health

ChenYiLins:
### Description

Standardize the ViewModels code, test the functional stability and determine the rationality of UX.


ChenYiLins:
## Code quality: About TimePage

### Description

- Change NumberBox to TextBox, Added support for localized separators, and supported `.` as decimal separator by default in all cases.
- Fixed a code error where OffsetTimeSettingsCardVisibility was not set.
- Add Apply Button to Geo Coordinates block.
- Add a time block to display the sunset and sunrise time when the geographic coordinates are customized.

### Supplement

@Jay-o-Way I need some help with this part of the UI change. In the previous UI changes, we forgot the sunrise and sunset times when displaying the custom geographic coordinates. Do we need to rearrange the positions of these components now?

### Screenshots

image

AutoDarkModeApp/Views/TimePage.xaml.cs
```diff
@@ -12,4 +13,35 @@ public TimePage()
         ViewModel = App.GetService<TimeViewModel>();
         InitializeComponent();
     }
+
+    private void CoordinatesTextBox_BeforeTextChanging(TextBox sender, TextBoxBeforeTextChangingEventArgs args)
+    {
+        var decimalSeparator = CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator;
+
+        args.Cancel = !args.NewText.All(c => char.IsDigit(c) || c == decimalSeparator[0] || c == '.' || c == '-');
```
Jay-o-Way, line 21: Looks like it will cancel when there's a "+" ?

AutoDarkModeApp/Views/TimePage.xaml
```diff
@@ -131,24 +156,24 @@
...
+                                    <TextBox
                                         MinWidth="120"
+                                        BeforeTextChanging="CoordinatesTextBox_BeforeTextChanging"
                                         Header="{helpers:ResourceString Name=TimeNumberBoxHeaderLon}"
-                                        Maximum="90"
-                                        Minimum="0"
-                                        Value="{x:Bind ViewModel.LonValue, Mode=TwoWay}" />
+                                        InputScope="Number"
+                                        Text="{x:Bind ViewModel.LonValue, Mode=TwoWay}" />
+                                    <Button
```
Jay-o-Way, line 171: If the binding is TwoWay, why do we need this button?

Jay-o-Way: Fun fact: notice the decimal
Comma period

ChenYiLins: > If the binding is TwoWay, why do we need this button?

This really should not be the expected behavior. Because TextBox can't update the value bound to the Text property in twoway when entering the Enter key, it will only update when it loses focus.

```c#
partial void OnLatValueChanged(string? value) => Debug.WriteLine("LatValue have changed");
```

image

AutoDarkModeApp/Views/TimePage.xaml.cs
ChenYiLins: `+` should be unnecessary. A fix has been submitted for the wrong input behavior.

AutoDarkModeApp/Views/TimePage.xaml
ChenYiLins: > Fun fact: notice the decimal

For the change of decimal point, hahahaha. I forget `CultureInfo.InvariantCulture`.

ChenYiLins:
> Sounds like that idea I had earlier? Show the **TimePicker** controls with every time-related setting, and IsEnabled=false for the 2 location settings?

I always think it's a good idea, and I'll try to realize it!

Jay-o-Way:
My opinion is i prefer to have all of them in view. Don't like scrollbars that much, for small areas.

ChenYiLins:
## Code quality: Using VariableSizedWrapGrid

### Description

Use VariableSizedWrapGrid instead of StackPanel using ScrollView. Just like the previous design.

AutoDarkModeApp/ViewModels/TimeViewModel.cs
```diff
@@ -108,37 +123,45 @@ public TimeViewModel(IErrorService errorService, ILocalSettingsService localSett
...
         DateTime nextUpdate = _builder.LocationData.LastUpdate.Add(_builder.Config.Location.PollingCooldownTimeSpan);
-        LocationHandler.GetSunTimesWithOffset(_builder, out DateTime SunriseWithOffset, out DateTime SunsetWithOffset);
         _dispatcherQueue.TryEnqueue(async () =>
         {
-            //TODO: A/B Testing. Let the time format completely follow the system settings
+            string timeFormat = CultureInfo.CurrentCulture.DateTimeFormat.ShortTimePattern;
+            bool isSystem12HourFormat = timeFormat.Contains('h');
+            TimePickHourClock = isSystem12HourFormat ? Windows.Globalization.ClockIdentifiers.TwelveHour : Windows.Globalization.ClockIdentifiers.TwentyFourHour;
+
             await LoadGeolocationData();
```
Jay-o-Way, line 146: Moving this is a good job. But still, now the _possible_ delay happens on navigating to the page. Still think it's better to move it to App Load. My wifi might be pretty slow, but it's a situation to be thinking about.
Delay

Jay-o-Way (changes requested):
One more change for even better performance please.
After that: is there more we want to work on in this PR?
CC @Armin2208 @Spiritreader

ChenYiLins:
> One more change for even better performance please. After that: is there more we want to work on in this PR?

What I expected was to improve the performance of all pages in this PR, and at the same time, It could include some UI changes and UX improvements. At the same time, the most important part is the migration of resx.

AutoDarkModeApp/ViewModels/TimeViewModel.cs
ChenYiLins: > Moving this is a good job. But still, now the _possible_ delay happens on **navigating to the page**. Still think it's better to move it to App Load. My wifi might be pretty slow, but it's a situation to be thinking about. Also caching and refreshing in the background is still an idea. But this can also be done in a new PR.

@Jay-o-Way Thank you very much for your test! Which is a serious problem, causing a very large page loading delay. However, after testing, this delay problem comes from LoadPostponeTimer. Regarding the display delay of TimePicker, can you test whether this problem exists in WPF version of ADM?

ChenYiLins: I have tried to solve the problem about the delay of UI change, because there is a big gap between calling UI refresh directly after RadioButton change and waiting for ConfigWatcher to detect it.

```
Ticks
Direct use: 8,810
Wait ConfigWatcher: 1,601,326
```

But for the problem of cache time, this is limited by the back end. Because before, switching the custom geographical location mode was not effective immediately, and there was only one set of data in the back end, both automatic geographical location detection and custom geographical location modes were saved in this set of data (the new data would replace the old data). This causes the delay visible to the naked eye in low-speed networks.

ChenYiLins:
> FYI: I tweaked a few small things around the **TokenizingTextBox**. I had not problem running it like this. If you or the others do, let me know. If so, we need to have a closer look at what the problem could be.

Thanks! Actually, I'm not familiar with TokenizingTextBox. I just found it when I wanted to change the UI. I tried to follow the example. There is nothing wrong with your change φ(*￣0￣) !

AutoDarkModeApp/ViewModels/TimeViewModel.cs
Jay-o-Way: Ohhh okay!

ChenYiLins:
## Regression: About SwitchModesPage

### Description

SwitchModesPage has restored all functions available.

## Feature: Process block list

### Description

TokenizingTextBox is enabled for Process block list.

### Screenshots

image

ChenYiLins:
## Code quality: About SystemAreasPage

### Description

- Solve Aot warning.
- Replace enum type.

ChenYiLins:
## Code quality: About SettingsPage

### Description

Thanks to @Jay-o-Way for your advice! SettingsPage has undergone some adjustments.

- Remove the setting of 12-hour clock.
- InfoBar was added for language change.
- ComboBox was added for update channel.

### Screenshots

image
image

ChenYiLins:
## Code quality: About AboutPage

### Description

Visual adjustment φ(*￣0￣) .

### Supplement

Whether we should still use clickable SettingsCard to point to files. Because in general, HyperlinkButton should be used as a link to a web page, which may cause misunderstanding to users because of the fatuity of various functions.

### Screenshots

#### Show the content center of gravity with color blocks

##### Old design

AboutPage design no 1

##### New design

AboutPage design no 2

#### Let's look at it

image

image

Jay-o-Way:
Chen, you are great! @Armin2208 if you have a minute? Would love to know what you think!

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
```diff
@@ -297,126 +298,158 @@ partial void OnIsAlwaysFullDwmRefreshChanging(bool value)
...
+            return;
+
         _builder.Config.Tunable.Trace = value;
-        _builder.Save();
+
+        SafeSaveBuilder();
     }
 
-    partial void OnLanguageChanged(string? value)
+    partial void OnLanguageChanged(string? oldValue, string? newValue)
```
Jay-o-Way, line 326: There's something funny going on here. When debugging, the number of references goes from 2 to 0. When searching the project for `OnLanguageChanged` I get zero results. When/where is this being called?
image

Jay-o-Way: Besides that,

- The "old value" should be the setting that the window is (still) using
- It doesn't open the info bar on the first change

ChenYiLins: Huh? I use the community version of VS, and it doesn't support displaying references. I tried to check the source generator, and it seems to be working normally. I tried to use debug output:

```c#
Debug.WriteLine("Old value: " + oldValue);
Debug.WriteLine("New value: "+ newValue);
```

I got these:

```
Old value: English (English)
New value: Français (French)
```

I don't know why it became a reference number of zero.

> Besides that, the "old value" should be the setting that the window is (still) using

Thank you for reminding me! It is true that I may keep changing languages and eventually return to the values still used in the current window, which will still trigger LanguageChanged. This is really not strict, I try to improve the code! ヾ(•ω•`)o

ChenYiLins:
## Regression: Localized language

### Description

All localized texts are back! ( •̀ ω •́ )✧ The tray can also switch languages normally, and some minor errors have been fixed.

ChenYiLins:
## Regression: Localized language

### Description

All localized texts are back! ( •̀ ω •́ )✧ The tray can also switch languages normally, and some minor errors have been fixed.

Jay-o-Way:
This is huge. Literally and figuratively 😄 Let's try to get this approved. And I suggest not focusing on AOT yet, don't know if the effort is worth it.

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
Jay-o-Way: I couldn't find the link between a UI event and this function. Can you help with that?

ChenYiLins:
> This is huge. Literally and figuratively 😄 Let's try to get this approved. And I suggest not focusing on AOT yet, don't know if the effort is worth it.

Ha ha ha ha, if nothing unexpected. Aot is troublesome (* ￣︿￣) , but it can significantly improve performance and reduce memory and storage occupation. I really don't think it should be handled in this PR. Let's focus on UI and functions.

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
ChenYiLins: [Here!](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/c712209bfb4a8442a09ea7b6c07b2a1baf26635d/AutoDarkModeApp/Views/SettingsPage.xaml#L50) `<ComboBox MinWidth="120" SelectedItem="{x:Bind ViewModel.Language, Mode=TwoWay}">`

Spiritreader:
> > This is huge. Literally and figuratively 😄 Let's try to get this approved. And I suggest not focusing on AOT yet, don't know if the effort is worth it.
> 
> Ha ha ha ha, if nothing unexpected. Aot is troublesome (* ￣︿￣) , but it can significantly improve performance and reduce memory and storage occupation. I really don't think it should be handled in this PR. Let's focus on UI and functions.

Changing publishing profiles needs to be tackled and approved by me first, otherwise the deployment system might break.
This should be a completely different PR.

In fact, I would suggest that PRs have a smaller scope from now on, that makes it easier to review, otherwise there's just too much.

Example:
- Changing something on the "Cursors" Page is one PR
- Changing something on the "Settings" page is another PR
- Changing something on the "About" page is another PR

Even if only a single UI element is changed.
Because after a while it becomes near impossible to keep track of everything that happens in the PR if you're not actively involved in making the changes.

AutoDarkModeApp/ViewModels/TimeViewModel.cs
Spiritreader: > I have tried to solve the problem about the delay of UI change, because there is a big gap between calling UI refresh directly after RadioButton change and waiting for ConfigWatcher to detect it.
> 
> ```
> Ticks
> Direct use: 8,810
> Wait ConfigWatcher: 1,601,326
> ```
> 
> But for the problem of cache time, this is limited by the back end. Because before, switching the custom geographical location mode was not effective immediately, and there was only one set of data in the back end, both automatic geographical location detection and custom geographical location modes were saved in this set of data (the new data would replace the old data). This causes the delay visible to the naked eye in low-speed networks.

The UI should never hang because it is waiting for data.
Winui3 should have asynchronous components that allow for the UI to load while other data is being fetched in the background.

The old WPF-based UI used asynchronous methods and an async dispatcher to prevent the UI thread from hanging. There must be something similar in Winui3, otherwise the UI framework would be pretty trash tbh.

ChenYiLins:
> Changing publishing profiles needs to be tackled and approved by me first, otherwise the deployment system might break.

Yes, we must prepare a new publishing profile for the new App later. We may consider PublishReadyToRun and PublishTrimmed.

> In fact, I would suggest that PRs have a smaller scope from now on, that makes it easier to review, otherwise there's just too 

Good idea, this PR is a bit huge now.

AutoDarkModeApp/ViewModels/TimeViewModel.cs
ChenYiLins: Yep. There is a similar mechanism in WinUI, because I didn't feel the delay because of the fast network before, （；´д｀）ゞ so I forgot to add a similar function, but it has been completed now.

Spiritreader (changes requested):
First of all, this looks great!

Some much needed improvements to the layouting and overall UI structure.

I agree that this should be merged asap, I've commented on a couple of issues that we maybe need to address!

Also apologies for the delays, time is scarce.

ChenYiLins:
> > Delay switching with a number of minutes #
> 
> sounds doesn't sound quite grammatically correct.
> 
> I would be "Delay switching by a number of minutes"
> 
> or we pick something else altogether, like "Move theme switch by a fixed minute offset" or sth.

In fact, I shouldn't be able to help us with the expression of the text (っ °Д °;)っ because I'm not a native English speaker. If you think there is anything to change, just change it (oﾟvﾟ)ノ

Spiritreader:
> because I'm not a native English speaker

You've done a wonderful job so far!
Also, it's probably good to get your feedback because the wording in Auto Dark Mode should be as easy to understand as possible!

Spiritreader:
I will merge this now,
However I have a question on the architectural side of localization, which I have now tracked in #965

Spiritreader (approved):

Issue #919, opened 2025-03-10
UI/UX: Unclear what the Offset settings do


Kiwow:
**Describe what issue or improvement you have with the usability design**

It's not clear what the "Light (minutes)" and "Dark (minutes)" values do in the Time > Offset settings. I opened the settings because I wanted the app to switch to Dark mode some time before sunset. It made sense to me that the Offset settings would provide what I need, but I was confused on how to do it. My best guess was that the numbers meant making the "Light" or "Dark" periods that many minutes longer.

I ended up asking Copilot about this – it explained how the setting works and pointed me to [this function in the source code](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/5922db4fd510722415cce40a3047b0b284f6e4d7/AutoDarkModeApp/Handlers/LocationHandler.cs#L64-L77) to support what it was saying. This is, in my opinion, too much to ask of a regular user to understand this feature.

**Describe how you would like to see it improved**

Add an explanation for what the "Light (minutes)" and "Dark (minutes)" values control – they offset the start of the "light" or "dark" period by the set number of minutes. A mention that the *numbers can be negative* would be nice as well.

I also thought it would be good to replace "Light" and "Dark" in this UI with "Sunrise" and "Sunset", but I understand that might not be accurate when a user has set custom hours for the switching, for instance.

Spiritreader:
Good point, at least a tooltip could provide more information, namely that Light (minutes) means "shift switch to light theme by x minutes" and Dark (minutes) means "shift switch to dark theme by x minutes")

Jay-o-Way:
Question: What is a reasonable maximum value? Currently it's arbitrarily max-ed as ±999 minutes, equals 16 hours and 23 minutes 😆

Armin2208:
> Question: What is a reasonable maximum value? Currently it's arbitrarily max-ed as ±999 minutes, equals 16 hours and 23 minutes 😆

it has no maximum value by design :D

ChenYiLins:
> Good point, at least a tooltip could provide more information, namely that Light (minutes) means "shift switch to light theme by x minutes" and Dark (minutes) means "shift switch to dark theme by x minutes")

Need to note: Should we add tool tip now? I think Jay's added string is a good illustration.

![Image](https://github.com/user-attachments/assets/4940429a-c8fd-4c41-bb62-152362791d63)

t0scale:
[...]

Issue #944, opened 2025-04-14
Migrate from Bing Maps to Azure Maps


Jay-o-Way:
Migrate from Bing Maps to Azure Maps:

- https://learn.microsoft.com/azure/azure-maps/migrate-bing-maps-overview
- source: https://github.com/microsoft/microsoft-ui-xaml/issues/10229#issuecomment-2540275280

Jay-o-Way:
I copied "Azure" from the comment(s), but if there's another simple server, that's also fine with me, of course.

Jay-o-Way:
_Today, Copilot gave me a few very simple PowerShell suggestions:_

Since you want automatic location detection in the background, we’ll tap into Windows’ built-in geolocation capabilities using PowerShell. Here's a clean and relatively lightweight approach:

### 🧭 Step 1: Automatically Fetch Location (Lat/Lon)
Windows exposes location data via the `System.Device.Location` namespace. You can access it in PowerShell like this:

```ps
Add-Type -AssemblyName System.Device

$GeoWatcher = New-Object System.Device.Location.GeoCoordinateWatcher
$GeoWatcher.Start()

while (($GeoWatcher.Status -ne 'Ready') -and ($GeoWatcher.Permission -ne 'Denied')) {
    Start-Sleep -Milliseconds 100
}

if ($GeoWatcher.Permission -eq 'Denied') {
    Write-Error 'Access Denied for Location Information'
} else {
    $location = $GeoWatcher.Position.Location
    $latitude = $location.Latitude
    $longitude = $location.Longitude
    Write-Output "Latitude: $latitude, Longitude: $longitude"
}
```

✅ This script runs silently and fetches your current coordinates. Make sure location services are enabled in Windows Settings → Privacy → Location.

### 🌅 Step 2: Get Sunrise/Sunset Times
Use the [Sunrise-Sunset API](https://sunrise-sunset.org/api) — it’s free and doesn’t require an API key. Here’s how to query it:

```ps
$apiUrl = "https://api.sunrise-sunset.org/json?lat=$latitude&lng=$longitude&formatted=0"
$response = Invoke-RestMethod -Uri $apiUrl

$sunriseUTC = [datetime]$response.results.sunrise
$sunsetUTC = [datetime]$response.results.sunset

# Convert to local time
$sunriseLocal = $sunriseUTC.ToLocalTime()
$sunsetLocal = $sunsetUTC.ToLocalTime()

[...]

PR #969, opened 2025-05-26
Code quality: About TimePage

ChenYiLins:
### Description

- Add the automatic saving of the offset, and use debounce to ensure the normal work of the configuration file.
- Limit the maximum value of offset, which needs more discussion. Do we need to set a reasonable maximum value, such as 24 hours? Because [it has no it has no maximum value by design now](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/issues/919#issuecomment-2764674271).
- Adjusted layout of some controls.

### DIFF
--- a/AutoDarkModeApp/ViewModels/TimeViewModel.cs
+++ b/AutoDarkModeApp/ViewModels/TimeViewModel.cs
@@ -17,6 +17,7 @@ public partial class TimeViewModel : ObservableRecipient
     private readonly AdmConfigBuilder _builder = AdmConfigBuilder.Instance();
     private readonly Microsoft.UI.Dispatching.DispatcherQueue _dispatcherQueue;
     private readonly IErrorService _errorService;
+    private readonly Microsoft.UI.Dispatching.DispatcherQueueTimer _debounceTimer;
     private bool _isInitializing;
 
     public enum TimeSourceMode
@@ -71,8 +72,6 @@ public partial class TimeViewModel : ObservableRecipient
     [ObservableProperty]
     public partial int OffsetDark { get; set; }
 
-    public ICommand SaveOffsetCommand { get; }
-
     [ObservableProperty]
     public partial bool IsPostponed { get; set; }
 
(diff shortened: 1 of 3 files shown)


Jay-o-Way:
Nit: in _debug_ there's no text next to this icon
image

ChenYiLins:
> Nit: in _debug_ there's no text next to this icon

Yes, it will work normally only when Svc is working normally.

Jay-o-Way:
### Adjusted the layout
Time settings

ChenYiLins:
> The city isn't filled in anymore, is this because of #944?

Maybe not? In my place, it is still working normally.

image

> Adjusted the layout

It looks pretty good.

ChenYiLins:
## Regression: Windows night mode can use Offset function

### Description

The Offset function can be used in Windows night mode. The maximum and minimum absolute value of Offset is set to 720 minutes (12 hours).

Jay-o-Way:
@Spiritreader I would like to re-do the "Postpone" part. Make clear what it does. Does it pause the app/svc or delay the switch? Can you help with this? Or a new PR for that?

Armin2208:
> @Spiritreader I would like to re-do the "Postpone" part. Make clear what it does. Does it pause the app/svc or delay the switch? Can you help with this? Or a new PR for that?

Could you please describe your goal more precisely? And could you please first show a concept, before making a commit? Thank you.
I prefer having a seperate PR for a design change in this case.

This is what Postpone does:
- Postpone **delays** theme switches by x hours or once (one entire day). 
  - Personally I would say, the user does not care about pausing or delaying, because for the user its kinda the same.
- Active delays are a collection of delays reported by the backend's postpone system (e.g if switching is delayed due to GPU usage, or an active program, or mouse activity)
- The delay button may either set a user-configurable postpone delay, or clear it, but does not remove non-user set postpones.

Jay-o-Way:
> Could you please describe your goal more precisely?

Just to update the UI. Make some room for explaining text. And each control should have a header. Also for Accessibility (Win Narrator) purposes.

Jay-o-Way (approved):

Issue #702, opened 2023-04-01
More user-friendly Hotkeys assignment User interface


oeloo:
Currently, assigning one of several hotkeys requires an unusual process that needs to be done in this order :
1. Turning off 'Enable system-wide hotkeys'
2. Entering the hotkeys you want in the user interface
3. Then turning on 'Enable system-wide hotkeys'
![image](https://user-images.githubusercontent.com/98819304/229270226-345530c3-2aa1-49ec-9278-82b8e3c24a69.png)

The problem is for any none Auto Dark Mode advanced user, it is not an usual way to specify hotkeys. In most apps (ex: Ditto, Flow Launcher), you just have to enter the hotkey, without having to fiddle with an on/off writing switch.

Here is are examples of user friendly hotkey setting interface: 

![image](https://user-images.githubusercontent.com/98819304/229270519-c967e494-8f55-4e35-812a-367994592e09.png)
This is taken from Flow Launcher, an other opensource application for Windows. As you can see there is no on/off writing switch to enter the hotkeys.

![image](https://user-images.githubusercontent.com/98819304/229270728-f8f27508-44a1-483d-b295-37044d54ebc0.png)
This is taken from Ditto, an other opensource application for Windows. There is a lot hotkeys that can be entered, and there is no on/off switch for entering these hotkeys.

Jay-o-Way:
The work on building the (settings) app with WinUI 3 is started. That will solve this.

ChenYiLins:
Now we have finished the design of the new hotkey interface. It will appear in Version 11 soon.

![image](https://github.com/user-attachments/assets/d22e0edb-ba70-4d33-b6d9-3dddbcd9c295)

![image](https://github.com/user-attachments/assets/4c128f35-27c6-47f9-a9b0-520f89d0a30b)

ChenYiLins:
> The problem is for any none Auto Dark Mode advanced user, it is not an usual way to specify hotkeys. In most apps (ex: Ditto, Flow Launcher), you just have to enter the hotkey, without having to fiddle with an on/off writing switch.

@Armin2208 @Jay-o-Way We need to discuss this. Do we need to disable hotkey editing when hotkeys are enabled? I am about to submit a PR on HotkeysPage, in which we will make a decision on this point.
My personal opinion does not need to be disabled, because first of all, the back end supports modifying the hotkey function when it is enabled, and I added the forced save button to make an extra guarantee for this function.

[...]

Issue #955, opened 2025-04-25
Several UI re-organisation suggestions


sjain882:
### 'Apps' & 'System' dropdowns do not reflect Windows naming ('App mode' & 'Windows mode')

Currently, it's like this:

![Image](https://github.com/user-attachments/assets/c93d4a97-654b-4aba-8fe4-531e7739a92e)

To remove ambiguity, rename them to exactly match the Windows naming ('App mode' and 'Windows mode' respectively).

Whether or not the Windows naming is optimal is irrelevant - each setting's name should accurately convey its effect.

Ideally, tooltips should be added to e.g, convey that 'Windows mode' reflects the system theme, such as the taskbar etc.

### 'Apps' & 'System' should not be under the 'Apps' section.

The setting title implies that you can control themes for individual apps here, but these settings only control global windows modes.

This section should be renamed to 'Theme Modes' or 'Theme Behaviour'.

### 'Dark Reader' & 'Recommended apps' should not be under the 'Apps' section.

These suggestions, while relevant to apps, are just vague suggestions, not actual app settings.

They should be moved to a separate section named 'Other'.

### 'Hotkeys' should not be under 'Switch Modes'.

This section's title is quite ambigious, but vaguely implies that you can control 'modes' (?) determining how the theme is switched.

Hotkeys should be moved to a separate section of their own.

This would only leave behind various switch conditions, removing the need for an ambigious section title. It can now be simply renamed to 'Conditions'.

When I started using this app, I spent a lot of time digging for various settings due to a combination of poor naming & organisation. I believe the above changes will solve this problem entirely.

I'm open to further suggestions for all of the above.

Jay-o-Way:
Good points, thank you!

1. I totally agree!
2. Known issue. We've started on a major re-work and these strings are already being changed.
- other points will be looked at

Armin2208:
Thank you very much for giving us suggestion on how to improve the UX. I really appreciate that.

1) I also agree with point 1, and we will change that for sure.
2) We will rename "Apps" to "System Areas" to better reflect what the settings do.
3) I actually think, this is the best place for recommending tools like Dark Reader. Because if I want to edit the areas, where the changes of Auto Dark Mode will reflect, I will look up at this page. In the future, we can maybe add a "web browser" collapsable box, where the dark reader recommendation sits in.
[...]

PR #970, opened 2025-05-26
Code quality: About SwitchModesPage and HotkeyPage

ChenYiLins:
### Description

- Limit GPU utilization to 100 (obviously it should not exceed 100 (っ °Д °;)っ).
- In non-battery mode, options related to power saving are no longer displayed (previously, SettingsCard was only disabled).
- Moved hotkeys to their own page and updated UX. #702

### DIFF
--- a/AutoDarkModeApp/Strings/en-us/Resources.resw
+++ b/AutoDarkModeApp/Strings/en-us/Resources.resw
@@ -925,7 +925,7 @@ Currently installed version: {0}, new version: {1}</value>
     <value>Save</value>
   </data>
   <data name="ForceSaveSettings" xml:space="preserve">
-    <value>Forced save</value>
+    <value>Force save</value>
   </data>
   <data name="Msg_HotkeyProblem" xml:space="preserve">
     <value>The hotkey didn't work? Don't worry, let's save it again</value>


Jay-o-Way:
Every line of `<Run Text="{helpers:ResourceString Name=Hotkeys}" />` is redundant an can be removed

Jay-o-Way:
I'll use a slider for the percentage.

ChenYiLins:
## Code quality: Add hotkey deletion method

### Description

Now we can use Esc, Backspace and Delete to clear the hotkey. At the same time, the description of the original SettingsCard is replaced, because now even if the hotkey is disabled, we can still edit it.

### Screenshots

image

ChenYiLins:
## Code quality: "SwitchModes" renamed to "Conditions"

### Description

Follow #955.

Jay-o-Way:
There's something strang about the hotkeys that are shown in the textbox. When I press <kbd>Ctrl</kbd> I see "Ctrl + Control". When I press <kbd>Ctrl</kbd>+<kbd>Alt</kbd> I see "Ctrl + Alt + Menu"...

Jay-o-Way:
A new icon for the NavMenu would be more appropriate now. Maybe a toggle switch?

ChenYiLins:
> There's something strang about the hotkeys that are shown in the textbox. When I press Ctrl I see "Ctrl + Control". When I press Ctrl+Alt I see "Ctrl + Alt + Menu"...

Thanks! I forgot to let some keys not return their values, and now it has been fixed.

ChenYiLins:
> A new icon for the NavMenu would be more appropriate now. Maybe a toggle switch?

I found a icon. Looks like "Control switching conditions (settings)"

image

ChenYiLins:
## Code quality: New Hotkeys page (Sorry to keep you waiting so long φ(*￣0￣) )

### Description

- Added a new page for managing hotkeys
- Adjust the page order (NavBar item)
- Added a new page to capture new hotkeys (refer to [PowerToys](https://github.com/microsoft/PowerToys)). Although the implementation is not very similar, ADM has its own hotkey saving method (saved in Yaml).

### Supplement

- There is still some work to be done, including the correction of strings
- Using custom controls to put the Content of Dialog is because of a Bug in WinUI. Dialog in Xaml has no animation and looks abnormal
- The way to refresh the hotkey of the current page is still not good (you need to refresh the whole page, you can consider using binding or other methods)

### Screenshots

image

image

gif

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
```diff
@@ -0,0 +1,127 @@
...
+            VirtualKey.Space => "Space",
+            VirtualKey.Tab => "Tab",
+            VirtualKey.Back => "Backspace",
+            VirtualKey.Delete => "Del",
+            VirtualKey.PageUp => "PgUp",
+            VirtualKey.PageDown => "PgDn",
+            VirtualKey.Left => "��",
+            VirtualKey.Right => "��",
+            VirtualKey.Up => "��",
+            VirtualKey.Down => "��",
```
Jay-o-Way: I'm guessing this happened because of different script?

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
```diff
@@ -0,0 +1,127 @@
...
+        var key = e.Key;
+
+        var isCtrl = IsKeyDown(VirtualKey.Control);
+        var isShift = IsKeyDown(VirtualKey.Shift);
+        var isAlt = IsKeyDown(VirtualKey.Menu);
+        var isWin = IsKeyDown(VirtualKey.LeftWindows) || IsKeyDown(VirtualKey.RightWindows);
+        var clearTextBox = IsKeyDown(VirtualKey.Escape) || IsKeyDown(VirtualKey.Back) || IsKeyDown(VirtualKey.Delete);
+
+        var hotkey = "";
+        hotkey += isCtrl ? "Ctrl + " : "";
```
Jay-o-Way: As @Armin2208 and @Spiritreader can vouch, in some languages these names/abbreviations are different. In German "Ctrl" = "Strng" (for Steuerung)
But honestly, I think we can get away with that for now 🤓

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
ChenYiLins: I forgot this. In fact, we don't need it. Just delete it.Tk u asking this question!

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
```diff
@@ -0,0 +1,127 @@
...
+        hotkey += isShift ? "Shift + " : "";
+        hotkey += isAlt ? "Alt + " : "";
+        hotkey += isWin ? "LWin + " : "";
```
Jay-o-Way: Is there a way to add these words as/to an array, and then use something like "ArrayToString" on it, with ` + ` as the delimiter?

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
ChenYiLins: At the same time, I would like to mention one more thing. In fact, we don't need anything in it now, because it doesn't fit with the back end. But this is user-friendly, for example, you will see the **"Enter"** key on the standard English keyboard instead of the **"Return"** displayed by the program.

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
```diff
@@ -0,0 +1,127 @@
...
+
+    private void StackPanel_PreviewKeyDown(object sender, Microsoft.UI.Xaml.Input.KeyRoutedEventArgs e)
+    {
+        var key = e.Key;
+
+        var isCtrl = IsKeyDown(VirtualKey.Control);
+        var isShift = IsKeyDown(VirtualKey.Shift);
+        var isAlt = IsKeyDown(VirtualKey.Menu);
+        var isWin = IsKeyDown(VirtualKey.LeftWindows) || IsKeyDown(VirtualKey.RightWindows);
+        var clearTextBox = IsKeyDown(VirtualKey.Escape) || IsKeyDown(VirtualKey.Back) || IsKeyDown(VirtualKey.Delete);
```
Jay-o-Way: NOTE: this doesn't work anymore now

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
ChenYiLins: Similar to what I mentioned above, I think it is user-friendly, but the working mode of the back-end now does not allow us to change it.

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
ChenYiLins: Sounds good (oﾟvﾟ)

Jay-o-Way: I got this one

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
ChenYiLins: Oh, I forgot to change this again. But since you have raised this question, do we still need to keep these clear keys such as **Esc** working? We already have a dialog button to clear **(reset)**.

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml
```diff
@@ -0,0 +1,45 @@
...
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:local="using:AutoDarkModeApp.UserControls"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d">
+
+    <StackPanel
+        Orientation="Vertical"
+        PreviewKeyDown="StackPanel_PreviewKeyDown"
+        Spacing="16">
+        <TextBlock Text="Press the key combination to change this shortcut key" />
```
Jay-o-Way: > Customers interact with products using different input methods: keyboard, mouse, touch, voice, and more. So use generic verbs that work with any input method. Don't use input-specific verbs, such as click or swipe. Instead, use the verbs listed here.

https://learn.microsoft.com/style-guide/procedures-instructions/describing-interactions-with-ui

I'll update this one.

Jay-o-Way:
> * Using custom controls to put the Content of Dialog is because of a Bug in WinUI. Dialog in Xaml has no animation and looks abnormal

Would you like to clarify this?

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
Jay-o-Way: Actually, we need to look at accessibility. Right now, you cannot <kbd>Tab</kbd> to those buttons (!)

ChenYiLins:
> Would you like to clarify this?

In fact, I didn't find a similar problem in WinUI's [Issues](https://github.com/microsoft/microsoft-ui-xaml/issues) (there was an Issue that lost animation, but it wasn't caused in Xaml). For comparison, you can check the dialog in ColorPickerPage.

#### Color Picker page Dialog

gif

#### Hotkeys page Dialog

gif

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
```diff
@@ -0,0 +1,130 @@
...
+    private static bool IsKeyDown(VirtualKey key)
+    {
+        var keyboardState = InputKeyboardSource.GetKeyStateForCurrentThread(key);
+        return keyboardState.HasFlag(CoreVirtualKeyStates.Down);
+    }
+
+    private static string GetKeyString(VirtualKey key)
+    {
+        if (key >= VirtualKey.A && key <= VirtualKey.Z)
+            return key.ToString().ToUpper();
```
Jay-o-Way: Here, intellisense suggests two things:
1. Specify current culture
   image
2. Use an invariant version
   image

I don't know which one is the best.
https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1311

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
```diff
@@ -0,0 +1,130 @@
...
+
+    private static bool IsKeyDown(VirtualKey key)
+    {
+        var keyboardState = InputKeyboardSource.GetKeyStateForCurrentThread(key);
+        return keyboardState.HasFlag(CoreVirtualKeyStates.Down);
+    }
+
+    private static string GetKeyString(VirtualKey key)
+    {
+        if (key >= VirtualKey.A && key <= VirtualKey.Z)
```
Jay-o-Way: How about non-Latin keyboards?

ChenYiLins: I believe there should be no problem, because it is **"VirtualKey"**, but it still needs to be tested.

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml.cs
ChenYiLins: In fact, this series of codes has been tested, which seems to be my redundant consideration. Sadly, I didn't take into account the real problems, such as the symbols **','** and **'.'** ... Let me study it

Jay-o-Way:
I see you referred a few bits of my code. For a reason? or merge-issue?

ChenYiLins:
> I see you reverted a few bits of my code. For a reason? or merge-issue?

It seems that there is a problem with the merger.

Jay-o-Way:
@ChenYiLins just curious: did you use the chat, linked in [this comment](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/issues/702#issuecomment-2922869298) for the code/design too? 🙂

ChenYiLins:
> just curious: did you use the chat, linked in [this comment](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/issues/702#issuecomment-2922869298) for the code/design too? 🙂

That's right! But the final design is not only realized by these words, I try to refer to the design in other software.

ChenYiLins:
### Summary

Awaiting review, @Jay-o-Way Do you have anything to add?

### Screenshots

image

image

Armin2208:
wow this is awesome !!!

ChenYiLins:
## Code quality: Force save hotkey

### Description

In the original, hotkeys can only be edited when they are disabled. Although this restriction has been removed in the new version, users can choose to force saving because of some accidents.

### Screenshots

image

Jay-o-Way:
- Well, that last new tooltip is **very** informal. What if there's another app that interferes with catching hotkeys? Doesn't a _click_ do the exact same thing as the automatic save? Will the user be notified if something goes wrong saving the settings? Is there any feedback if the action is succesful?
  - Also, the button makes it look like you **need** to press it before anything is saved; not as an "extra".
- Looks like I can't add any app to the block-list? It's a fixed list?
- @Armin2208 or @Spiritreader need to judge if the issue with the GPU percentage is acceptable

ChenYiLins:
> * Well, that last new tooltip is **very** informal. What if there's another app that interferes with catching hotkeys? Doesn't a _click_ do the exact same thing as the automatic save? Will the user be notified if something goes wrong saving the settings? Is there any feedback if the action is succesful?

For interfering with the program to get the hotkey, it seems that the direct result is that any hotkey will not be displayed.
Click Save to avoid the possibility of saving failure when the hotkey function is enabled. Although I have never met it, I believe that there was a reason let us don't  edit hotkeys when the hotkey funcition is enabled before? )

>   * Also, the button makes it look like you **need** to press it before anything is saved; not as an "extra".

Indeed, I forgot this.

> * Looks like I can't add any app to the block-list? It's a fixed list?

Block list only lists the currently running window programs, but you can join them by entering their own program names.
However, there does seem to be a problem now. When entering a name, spaces will be automatically separated.

image

AutoDarkModeApp/Views/HotkeysPage.xaml
```diff
@@ -0,0 +1,79 @@
...
+                    ScrollViewer.VerticalScrollMode="Disabled"
+                    SelectionMode="None">
+                    <ListView.ItemTemplate>
+                        <DataTemplate x:DataType="local:HotkeysDataObject">
+                            <controls:SettingsCard Header="{x:Bind Name}">
+                                <StackPanel Orientation="Horizontal" Spacing="8">
+                                    <TextBox
+                                        MinWidth="160"
+                                        IsReadOnly="True"
+                                        Text="{x:Bind Keys, Mode=OneWay}" />
```
Jay-o-Way, line 48: This name `Keys` should be a little bit more specific. Note that there's also a list with this name.

Armin2208:
>     * @Armin2208 or @Spiritreader need to judge if the issue with the GPU percentage is acceptable

There is no issue with GPU percentage. The state is acceptable and good. I don't want to be like Microsoft, just capping the utilization of everything at 100% and because of this having complete meaningless numbers in Task Manager.
So: GPU utilization can exceed 100%, the textbox accepts values over 100% and this is just how things work today. So no need to change here something.

Armin2208:
> Looks like I can't add any app to the block-list? It's a fixed list?

How about changing the text to: "Start typing to add running apps".
Also: can't we display the list of all running apps then just clicking on the box? Or is typing necessary?

AutoDarkModeApp/Views/HotkeysPage.xaml
ChenYiLins: This should be the naming error of List. As a member property of a type, Keys will not conflict with external types with the same name.

AutoDarkModeApp/Views/HotkeysPage.xaml
```diff
@@ -15,13 +15,13 @@
 
                 <!--  Hotkeys  -->
                 <Grid>
-                    <TextBlock Style="{StaticResource SettingsSectionHeaderTextBlockStyle}" Text="{helpers:ResourceString Name=Hotkeys}" />
```
ChenYiLins: Should we keep this TextBlock? I can understand that the whole part of this page corresponds to "Hotkeys", so we don't need TextBlock to explain it. But the buttons on the right look visually unbalanced.


HotkeysPage design no 1

ChenYiLins:
> There is no issue with GPU percentage. The state is acceptable and good. I don't want to be like Microsoft, just capping the utilization of everything at 100% and because of this having complete meaningless numbers in Task Manager. So: GPU utilization can exceed 100%, the textbox accepts values over 100% and this is just how things work today. So no need to change here something.

The strange thing is that in [previous versions of ADM](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/8daffc4006f3d2ed2eb000207adaacc5d165dd35/AutoDarkModeApp/Pages/PageSwitchModes.xaml#L84), there was a limit to the maximum value of 100. Does the back-end support value exceed 100%?

ChenYiLins:
## Code quality: More visual indication

### Description

Change the button text to let the user know that it is not necessary to click every time.There is a clear visual effect when saving successfully or failing.

### Screenshots

image

gif

Jay-o-Way:
Automation property (Name) is needed for Narrator. Please add back 🙂

Jay-o-Way (approved):

ChenYiLins:
> How about changing the text to: "Start typing to add running apps". Also: can't we display the list of all running apps then just clicking on the box? Or is typing necessary?

We have a list that shows running programs, similar to the previous ADM. Typing is optional.

image

AutoDarkModeApp/Views/HotkeysPage.xaml
```diff
@@ -0,0 +1,75 @@
...
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d">
+
+    <ScrollViewer>
+        <Grid x:Name="ContentArea" Margin="{StaticResource NavigationViewPageContentMargin}">
+            <StackPanel Style="{StaticResource PageBaseStackPanelStyle}">
+
+                <!--  Hotkeys  -->
+                <Grid>
+                    <TextBlock Style="{StaticResource SettingsSectionHeaderTextBlockStyle}" Text="{helpers:ResourceString Name=Hotkeys}" />
```
Jay-o-Way: I dont want this. It is the same as the page title and there is only one group.

ChenYiLins: Actually, I referred to Files for the design of this button. There is also only one group, but only for visual balance.

Or there is a compromise, wrap this button in SettingsCard, so we don't need ToolTip.

image

Jay-o-Way: A settingscard or a (neutral) infobar would be good too (or better) , because it has space for some explaining text, as you say. Also helps with Narrator.

Armin2208:
> Change the button text to let the user know that it is not necessary to click every time.There is a clear visual effect when saving successfully or failing.

Actually, I think the "forced preservation" is a bit too technical. First, it's in the wrong time form. It must be "force preservation", because it did not happen before you click the button. "Forced" alreay happened.
Maybe "Force save" is better.
Or just "save".
The button is only a backup nayways, you can add the save button at the bottom of the page as a new box of the ui, blending into the rest. Discovery is then not the greatest, but its clear and its there, not relying on a mouse hover description.
Because I whish to have the Auto Dark Mode UI less technical.

Armin2208:
> > There is no issue with GPU percentage. The state is acceptable and good. I don't want to be like Microsoft, just capping the utilization of everything at 100% and because of this having complete meaningless numbers in Task Manager. So: GPU utilization can exceed 100%, the textbox accepts values over 100% and this is just how things work today. So no need to change here something.
> 
> The strange thing is that in [previous versions of ADM](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/8daffc4006f3d2ed2eb000207adaacc5d165dd35/AutoDarkModeApp/Pages/PageSwitchModes.xaml#L84), there was a limit to the maximum value of 100. Does the back-end support value exceed 100%?

Sam already replied to that: https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/970#issuecomment-2917248819

ChenYiLins:
> Actually, I think the "forced preservation" is a bit too technical. 

At first, I wanted users to think that this is an alternate option button (you don't need to use it every time).

> Because I whish to have the Auto Dark Mode UI less technical.

But I also think this is a very important point. We should not make some explanations that could have been obvious obscure and complicated.

Jay-o-Way:
@ChenYiLins "forced" is passed tense (or "past participle"), is this intentional?

ChenYiLins:
> "forced" is passed tense, is this intentional?

Uh!! Sorry. I forget this.

Jay-o-Way:
Okay unless @Armin2208 or @Spiritreader have any objections, I'm willing to merge this

PR #971, opened 2025-05-26
Code quality: About ThemePickerPage

ChenYiLins:
### Description

- Fixed an exception caused by the theme not loading normally.
- Adjusted text (tutorial and hint)

### DIFF
--- a/AutoDarkModeApp/Strings/en-us/Resources.resw
+++ b/AutoDarkModeApp/Strings/en-us/Resources.resw
@@ -889,6 +889,14 @@ Currently installed version: {0}, new version: {1}</value>
   <data name="WindowsAutostartDisabled" xml:space="preserve">
     <value>You have disabled the auto start entry via Windows.</value>
   </data>
+  <data name="ThemeTutorialDescription" xml:space="preserve">
+    <value>1. Go to Windows Settings &gt; Personalization &gt; Colors.
+2. Change the system color to 'Light' for the light theme.
+3. Go to Windows Settings &gt; Personalization &gt; Themes.
+4. Select your favorite wallpaper, mouse pointers and accent color.
+5. Save your theme.
+6. Repeat this process for the dark theme.</value>
+  </data>
   <data name="Beta" xml:space="preserve">
     <value>Beta</value>
   </data>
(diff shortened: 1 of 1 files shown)


Jay-o-Way:
Ideas:

- personally, I would use a different control to present the "create your themes" steps. Probably a **Rich TextBlock**.
- I think, in general, an **InfoBar** is mostly for _temporary_ situations. So, the text "This option does not support default themes" could also be in a **TextBlock** with a simple, colored ℹ️ icon before it. Compare in Windows Settings:
image

ChenYiLins:
> as you said earlier (somewhere) it looks like there is an issue with **NavigateUri** and `ms-settings`. I wonder why...

This is actually a [Bug](https://github.com/microsoft/microsoft-ui-xaml/issues/10428) brought by WinUI. We can try to use Click. Simply put, a window should have popped up to verify the jump event, but ... Hahaha

ChenYiLins:
> detail: now the text is inside a Card, and there is a visual glitch: it doesn't wrap and (the clickable area of) the hyperlinks are stretched. Using `SettingsCard.Description` solves the wrapping, not the stretching, but uses a different font style... (the stretching issue is probably not related to the settingscard)

About the wrapping, just because we forgot `ContentAlignment="Left"`.
gif

For _"the hyperlinks are stretched"_. Actually, I don't quite understand what you mean. We use Hyperlink to wrap the whole string, so we have a long clickable area.

```xaml
<Hyperlink Click="SettingsColorsHyperlink_Click">
         <Run Text="{helpers:ResourceString Name=ThemeTutorial_GoColors}" />
</Hyperlink>
```

Jay-o-Way:
ContentAlignment: I tried that, but it didn't work for me 🤔🤷🏻‍♂️

Hyperlinks: Have a look at the mouse pointer in the gif.

I've simplified the text for now.

Armin2208:
I liked the numbering more than the bullet points, because they gave a sense of step 1, step 2 etc.

ChenYiLins:
@Jay-o-Way Do we decide to join Hyperlink or not? Because it is easy to jump in GameBar settings (through a simple hotkey). But this is not the case for our current situation (although it does not seem particularly complicated, users can choose to look on the left and do it on the right)

ChenYiLins:
> Anything else we can add to this PR?

See what you all think, I think it's almost done? (oﾟvﾟ)ノ

Jay-o-Way (approved):
LGTM

PR #981, opened 2025-06-21
Code quality: About SystemAreasPage

ChenYiLins:
### Description

- Change the text to match Windows settings #955 
- Redesign the display logic of SettingCard, that is, hide it directly when it is disabled instead of displaying a completely inoperable SettingCard
- Add HyperlinkButton link to Windows settings -> Personalization -> Colors
- Fix the code error, the theme will not change immediately after changing the settings

### DIFF
--- a/AutoDarkModeApp/Strings/en-us/Resources.resw
+++ b/AutoDarkModeApp/Strings/en-us/Resources.resw
@@ -942,4 +942,13 @@ Currently installed version: {0}, new version: {1}</value>
   <data name="AddApps" xml:space="preserve">
     <value>Start typing to add any apps</value>
   </data>
+  <data name="AppMode" xml:space="preserve">
+    <value>App mode</value>
+  </data>
+  <data name="WindowsMode" xml:space="preserve">
+    <value>Windows mode</value>
+  </data>
+  <data name="WindowsColorsSetting" xml:space="preserve">
+    <value>Personalized colors setting of Windows</value>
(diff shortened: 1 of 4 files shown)


AutoDarkModeApp/Strings/en-us/Resources.resw
```diff
@@ -942,4 +942,13 @@ Currently installed version: {0}, new version: {1}</value>
   <data name="AddApps" xml:space="preserve">
     <value>Start typing to add any apps</value>
   </data>
+  <data name="AppMode" xml:space="preserve">
+    <value>App mode</value>
```
Jay-o-Way, line 946: Hm, I though I had already updated these thing in a previous PR. Maybe there was a merge issue then?

ChenYiLins: It maybe ignored in a conflicted merger.

Spiritreader:
wonderful, thank you <3

> The theme will not change immediately after changing the settings

Whether we want to instantly change the theme after modifying a setting depends a little on what the user expects to happen when a setting is changed, but we'll see if there needs to be more adjustment

ChenYiLins:
> Whether we want to instantly change the theme after modifying a setting depends a little on what the user expects to happen when a setting is changed, but we'll see if there needs to be more adjustment

In this page, I think it is more intuitive to change it immediately.

PR #984, opened 2025-06-26
Code quality: About ColorizationPage

ChenYiLins:
## Description

- Add new UserControl to implement ColorPicker
- Adding the title of the dialog looks more in line with Microsoft's development guidelines
- Fix that SettingsCard keeps expanding
- Add localized references

## Screenshots

image

### DIFF
--- a/AutoDarkModeApp/UserControls/ColorPickerDialogContentControl.xaml
+++ b/AutoDarkModeApp/UserControls/ColorPickerDialogContentControl.xaml
@@ -8,8 +8,10 @@
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     mc:Ignorable="d">
 
-    <StackPanel>
-        <ColorPicker ColorChanged="SelectColorColorPicker_ColorChanged" Color="{x:Bind CustomColor}" />
-    </StackPanel>
+    <ScrollViewer>
+        <StackPanel Padding="16">
+            <ColorPicker ColorChanged="SelectColorColorPicker_ColorChanged" Color="{x:Bind CustomColor}" />
+        </StackPanel>
+    </ScrollViewer>
 
 </UserControl>
--- a/AutoDarkModeApp/UserControls/ColorPickerDialogContentControl.xaml
+++ b/AutoDarkModeApp/UserControls/ColorPickerDialogContentControl.xaml
@@ -1,0 +1,15 @@
+<UserControl
+    x:Class="AutoDarkModeApp.UserControls.ColorPickerDialogContentControl"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:helpers="using:AutoDarkModeApp.Helpers"
+    xmlns:local="using:AutoDarkModeApp.UserControls"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d">
+
+    <StackPanel>
+        <ColorPicker ColorChanged="SelectColorColorPicker_ColorChanged" Color="{x:Bind CustomColor}" />
+    </StackPanel>
+
+</UserControl>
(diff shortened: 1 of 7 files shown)


AutoDarkModeApp/UserControls/ColorPickerDialogContentControl.xaml
Jay-o-Way, line 1: In the future I **really** want to investigate why we think this is needed and see if we can bring it back to "standard" dialog creation. Throughout the project.

ChenYiLins: This problem has existed for quite a long time, and Microsoft seems to have been focusing on the update of AI functions. Start using WinUI for task manager with Win11. Until now, in the latest 24H2 system, the dialog box of task manager ending the task still loses the open animation effect.

ContentDialog has many unimaginable bugs:

- https://github.com/microsoft/microsoft-ui-xaml/issues/8476
- https://github.com/microsoft/microsoft-ui-xaml/issues/8402
- https://github.com/microsoft/microsoft-ui-xaml/issues/10215
- https://github.com/microsoft/microsoft-ui-xaml/issues/8640
- https://github.com/microsoft/microsoft-ui-xaml/issues/6577

AutoDarkModeApp/Views/ColorizationPage.xaml
```diff
@@ -23,10 +23,7 @@
                 </controls:SettingsCard>
 
                 <!--  Light  -->
-                <controls:SettingsExpander
-                    Header="{helpers:ResourceString Name=LightTheme}"
-                    HeaderIcon="{ui:FontIcon Glyph=&#xE706;}"
-                    IsExpanded="True">
```
Jay-o-Way: Personal opinion: I don't like the need to do extra "clicks" for so little controls. I saw you used this "collapsed" approach also on the **About** page - and a few other places. Can you explain why you prefer this?

If you look at Windows Settings, there are many expanders that are open and many that are closed. It depends on the content. Let's find a user friendly balance in this app :)

Jay-o-Way (changes requested):
small thing

AutoDarkModeApp/Views/ColorizationPage.xaml
ChenYiLins: Let's keep it open. Regarding whether to expand or not, I think it depends on whether the user needs to use this part of the content for the first time when opening this page (of course, it is a mistake to collapse it in this page)

I noticed a [discussion](https://github.com/CommunityToolkit/Windows/discussions/680#discussion-8274296) that you also participated in in the CommunityToolkit. In fact, I like the idea very much.

Jay-o-Way (changes requested):
Gotta fix the Dialog window: it is constrained to parent window size, which can be too small to show content.
image
https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.primitives.popup.shouldconstraintorootbounds?view=winrt-26100 doesn't work here 😒

We can
- change to a **Flyout** (e.g. PowerToys) or
- add scrolling in the content (like Windows Settings)

ChenYiLins:
> Gotta fix the Dialog window: it is constrained to parent window size, which can be too small to show content. https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.primitives.popup.shouldconstraintorootbounds?view=winrt-26100 doesn't work here 😒
> 
> We can
> 
> * change to a **Flyout** (e.g. PowerToys) or
> * add scrolling in the content (like Windows Settings)

Huh? I guess it's another annoying bug Let's use the scroll.

Jay-o-Way (approved):

Spiritreader:
The publish profiles were updated to push the beta out, please update the other project dependencies (Shell, Comms, Svc, Lib) because currently this PR overwrites those changes.

Should be fixed by pulling the AutoDarkMode winui3 branch into your fork.

ChenYiLins:
> The publish profiles were updated to push the beta out, please update the other project dependencies (Shell, Comms, Svc, Lib) because currently this PR overwrites those changes.
> 
> Should be fixed by pulling the AutoDarkMode winui3 branch into your fork.

Done.

ChenYiLins:
Transfer to: #1001.

PR #1003, opened 2025-07-21
Update error message localization keys

Jay-o-Way:
**errorOcurredTitle** and **ErrorMessageBox** were old and created a crash.


ChenYiLins (approved):
LGTM φ(*￣0￣)

PR #1004, opened 2025-07-21
Improvements to the code (demo only)

Jay-o-Way:
Mostly a "proof of concept". Mostly refactoring code. Should be no functional changes.

### DIFF
--- a/AutoDarkModeApp/Strings/en-us/Resources.resw
+++ b/AutoDarkModeApp/Strings/en-us/Resources.resw
@@ -230,7 +230,7 @@ On rare occasions, Windows may emit a beeping sound when the mouse is moved duri
     <value>Color filter (sends Win + Ctrl + C)</value>
   </data>
   <data name="ColorizationPick_Description" xml:space="preserve">
-    <value>Customize the Windows system color scheme to your liking.</value>
+    <value>Customize the Windows system color scheme to your liking</value>
   </data>
   <data name="Conditions" xml:space="preserve">
     <value>Conditions</value>
@@ -956,5 +956,11 @@ Currently installed version: {0}, new version: {1}</value>
   </data>
   <data name="WindowsColorsSetting" xml:space="preserve">
     <value>Personalized colors setting of Windows</value>
+  </data>
+  <data name="SelectColor" xml:space="preserve">
+    <value>Select color</value>
+  </data>
+  <data name="ChooseCustomColor" xml:space="preserve">
+    <value>Choose a custom color</value>
   </data>
 </root>
--- a/AutoDarkModeApp/UserControls/ColorPickerDialogContentControl.xaml
+++ b/AutoDarkModeApp/UserControls/ColorPickerDialogContentControl.xaml
@@ -1,0 +1,17 @@
+<UserControl
+    x:Class="AutoDarkModeApp.UserControls.ColorPickerDialogContentControl"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:helpers="using:AutoDarkModeApp.Helpers"
+    xmlns:local="using:AutoDarkModeApp.UserControls"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d">
+
+    <ScrollViewer>
+        <StackPanel Padding="16">
+            <ColorPicker ColorChanged="SelectColorColorPicker_ColorChanged" Color="{x:Bind CustomColor}" />
+        </StackPanel>
+    </ScrollViewer>
+
+</UserControl>
--- a/AutoDarkModeApp/UserControls/ColorPickerDialogContentControl.xaml.cs
+++ b/AutoDarkModeApp/UserControls/ColorPickerDialogContentControl.xaml.cs
@@ -1,0 +1,32 @@
+using Microsoft.UI;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Windows.UI;
+
+namespace AutoDarkModeApp.UserControls;
+
+public sealed partial class ColorPickerDialogContentControl : UserControl
+{
+    public Color CustomColor
+    {
+        get => (Color)GetValue(CustomColorProperty);
+        set => SetValue(CustomColorProperty, value);
+    }
+
+    public static readonly DependencyProperty CustomColorProperty = DependencyProperty.Register(
+        "CustomColor",
+        typeof(Color),
+        typeof(ColorPickerDialogContentControl),
+        new PropertyMetadata(Colors.White)
+    );
+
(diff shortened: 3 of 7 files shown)


AutoDarkModeApp/ViewModels/SystemAreasViewModel.cs
```diff
@@ -240,15 +243,16 @@ partial void OnSystemSwitchComponentModeChanged(SystemSwitchMode value)
         }
         AccentColorForTaskbarSettingsCardVisible =
             SystemSwitchComponentMode != SystemSwitchMode.AlwaysLight && SystemSwitchComponentMode != SystemSwitchMode.AccentOnly && SystemSwitchComponentMode != SystemSwitchMode.Disabled;
-        IsAdaptiveTaskbarAccent = value == SystemSwitchMode.AccentOnly;
+        IsAdaptiveTaskbarAccent = value
+            == SystemSwitchMode.AccentOnly;
```
Jay-o-Way, line 247:
```suggestion
        IsAdaptiveTaskbarAccent =
        value == SystemSwitchMode.AccentOnly;
```

Spiritreader:
Do we have tests written to verify that these code changes to not modify the behavior?

Jay-o-Way:
I dont know about tests. Is there something specific you are worried for?

AutoDarkModeApp/ViewModels/SystemAreasViewModel.cs
```diff
@@ -9,6 +9,7 @@ namespace AutoDarkModeApp.ViewModels;
 
 public partial class SystemAreasViewModel : ObservableRecipient
 {
+    private const string Source = "SystemAreasViewModel";
```
Jay-o-Way: I see both "Location" and "Source" are used. I think "location" is better?

Jay-o-Way:
Question: can `allowshowdisplay` ever be **true**?

```cs
//Service.cs

internal class Service : Form
{
    private readonly bool allowshowdisplay = false;
...

    // line 143
    protected override void SetVisibleCore(bool value)
    {
        base.SetVisibleCore(allowshowdisplay ? value : allowshowdisplay);
    }
```

Jay-o-Way:
### in Service.cs

- What's the difference between these two?

```cs
    private AdmConfigBuilder Builder { get; } = AdmConfigBuilder.Instance();
    // ...
    private readonly AdmConfigBuilder builder = AdmConfigBuilder.Instance();
```

- This is suggested:

image

> CA1859: Use concrete types when possible for improved performance

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1859

- Can `allowshowdisplay` ever be **true**?

```cs
internal class Service : Form
{
    private readonly bool allowshowdisplay = false;
    // ...

    // line 143
    protected override void SetVisibleCore(bool value)
    {
        base.SetVisibleCore(allowshowdisplay ? value : allowshowdisplay);
    }
```

AutoDarkModeSvc/Core/ThemeManager.cs
```diff
@@ -209,13 +216,18 @@ public static void UpdateTheme(SwitchEventArgs e)
...
-                GlobalState.Instance().RefreshThemes(AdmConfigBuilder.Instance().Config);
-                themeModeNeedsUpdate = ThemeHandler.ThemeModeNeedsUpdate(newTheme);
+                case SwitchSource.SystemUnlock:
+                case SwitchSource.SystemResume:
+                    // check for theme changes on system unlock or resume
+                    GlobalState.Instance().RefreshThemes(AdmConfigBuilder.Instance().Config);
+                    themeModeNeedsUpdate = ThemeHandler.ThemeModeNeedsUpdate(newTheme);
+                    break;
+                default:
+                    themeModeNeedsUpdate = ThemeHandler.ThemeModeNeedsUpdate(newTheme);
```
Jay-o-Way, line 228: Still redundant?

AutoDarkModeSvc/Handlers/WallpaperHandler.cs
```diff
@@ -105,24 +109,28 @@ public static void SetWallpapers(List<MonitorSettings> monitorSettings, Wallpape
...
-                    else
-                    {
-                        handler.SetWallpaper(monitor.DeviceId, monitorSetting.DarkThemeWallpaper);
-                    }
-                }
-                else
-                {
-                    if (!File.Exists(monitorSetting.LightThemeWallpaper))
-                    {
-                        Logger.Warn($"wallpaper does not exist. path ${monitorSetting.DarkThemeWallpaper}, monitor ${monitor.DeviceId}");
```
Jay-o-Way, line 123: Looks like we found a typo here

AutoDarkModeApp/ViewModels/SystemAreasViewModel.cs
ChenYiLins: I personally prefer Location, and let's consider whether it is necessary to make the error message more detailed.

ChenYiLins:
> ### in Service.cs
> * What's the difference between these two?
> 
> ```cs
>     private AdmConfigBuilder Builder { get; } = AdmConfigBuilder.Instance();
>     // ...
>     private readonly AdmConfigBuilder builder = AdmConfigBuilder.Instance();
> ```

There seems to be little difference in the final effect, one is attribute and the other is field. There is little difference in performance, but using attributes seems to be a more standardized approach (preparing for possible future updates)

> * This is suggested:
> 
> > CA1859: Use concrete types when possible for improved performance
> 
> https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1859

Sure. This is a good suggestion.

> * Can `allowshowdisplay` ever be **true**?
> 
> ```cs
> internal class Service : Form
> {
>     private readonly bool allowshowdisplay = false;
>     // ...
> 
>     // line 143
>     protected override void SetVisibleCore(bool value)
>     {
>         base.SetVisibleCore(allowshowdisplay ? value : allowshowdisplay);
>     }
> ```

No, it will never be changed.

AutoDarkModeSvc/Core/ThemeManager.cs
ChenYiLins: Ah ... maybe it's standard practice?

AutoDarkModeSvc/Handlers/WallpaperHandler.cs
ChenYiLins: Yep.

Spiritreader:
> * 
> 
> 
> ```cs
>     private AdmConfigBuilder Builder { get; } = AdmConfigBuilder.Instance();
>     // ...
>     private readonly AdmConfigBuilder builder = AdmConfigBuilder.Instance();
> ```

That is a really bad oversight, the lowercase builder should not be used.



> * This is suggested:
> 
> 
> image
> 
> > CA1859: Use concrete types when possible for improved performance

The reason why the concrete type is not used here was because ADM used to swap messaging backends in the past quite often until a stable one was found, and still has the capability to use a backend that is not based on namedpipes. Making it a concrete type would increase refactoring efforts when modifying the messaging backend. There aren't really any high performance or energy demaning operations being called here, so this warning can be disregarded.

PR #1011, opened 2025-08-06
Create dotnet-desktop.yml

ReuelAlbert-Dev:
deskop application compability cross device

### DIFF
--- a/.github/workflows/dotnet-desktop.yml
+++ b/.github/workflows/dotnet-desktop.yml
@@ -1,0 +1,115 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+# This workflow will build, test, sign and package a WPF or Windows Forms desktop application
+# built on .NET Core.
+# To learn how to migrate your existing application to .NET Core,
+# refer to https://docs.microsoft.com/en-us/dotnet/desktop-wpf/migration/convert-project-from-net-framework
+#
+# To configure this workflow:
+#
+# 1. Configure environment variables
+# GitHub sets default environment variables for every workflow run.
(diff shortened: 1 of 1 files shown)


Jay-o-Way:
Hi there. Since there is no issue for this PR, can you explain exactly what this change means and why it would be needed?

ReuelAlbert-Dev:
issue reduce background task in any desk application

Jay-o-Way:
@ReuelAlbert-Dev you're going to have to give us more than that. Besides, the info mentions WPF and we are not using that in the new (beta) version.

ReuelAlbert-Dev:
Clarify whether the workflow is intended for WPF apps or if it’s framework-agnostic

Spiritreader:
It appears as if you're trying to add a github workfllow for automatically building Auto Dark Mode. Which in theory would be a cool thing to have, but since we distribute via InnoSetup not somthing that can be done with quite a bit of refactoring.

While our deployment system (if you can even call it that) could definitely use some improvements, it requires more discussion at a laster point in time (post ADM 11 release).

The template you provided doesn't seem to actually be tailored towards ADM in any way or form, it doesn't even specify the solution name:

```
 env:
      Solution_Name: your-solution-name                         # Replace with your solution name, i.e. MyWpfApp.sln.
      Test_Project_Path: your-test-project-path                 # Replace with the path to your test project, i.e. MyWpfApp.Tests\MyWpfApp.Tests.csproj.
      Wap_Project_Directory: your-wap-project-directory-name    # Replace with the Wap project directory relative to the solution, i.e. MyWpfApp.Package.
      Wap_Project_Path: your-wap-project-path                   # Replace with the path to your Wap project, i.e. MyWpf.App.Package\MyWpfApp.Package.wapproj.
```

If you would like to contribute, feel free to move your inquiry to the discussions page with more detail, and please give us more respect than talking to us like we're LLMs.

PR #1016, opened 2025-08-18
remove unused implementation

### LINKED ISSUE #1015: Cleanup: WallpaperType.All
in file _WallpaperSwitch.cs_ under `HandleSwitchByType` under `WallpaperType.All` we see
```cs
Logger.Error("not implemented anymore");
```

But in the rest of the file WallpaperType.All is still being handled...

> Is being phased out in favor of the multi monitor wallpaper api
> Handle switch by type is the start of the chain and the rest of the chain is still there in case it was still needed b, but since we dropped support for w10 versions lower than  21H2 this is no longe required

~ Sam

krisbiradar:
closes  #1015

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -3,7 +3,7 @@
 github: ['Spiritreader']
 patreon: # Replace with a single Patreon username
 open_collective: # Replace with a single Open Collective username
-ko_fi: spiritreader
+ko_fi: # Replace with a single Ko-Fi username
 tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
 community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
 liberapay: # Replace with a single Liberapay username
--- a/AutoDarkModeApp/Pages/PageDonation.xaml
+++ b/AutoDarkModeApp/Pages/PageDonation.xaml
@@ -136,35 +136,35 @@
         </StackPanel>
 
         <!-- Kofi Donation Button -->
-        <Image 
+        <!--
+        <Image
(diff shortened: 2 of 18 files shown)


Jay-o-Way:
Thanks again for starting this. There are multiple places of "WallpaperType.All" in this file. Also, we should check if there is any code in other files that leverage this thing, like an **enum**.

Spiritreader:
I've talked with @Jay-o-Way about the scope of it briefly, but the gist is that the StateUpdate, TypeNeedsUpdate and Logger functions in the WallpaperSwitch.cs also need to be removed.

This isn't the only place where the wallpaper type is referenced, there's also enums for the wallpaper type.
For those enums, an argument can be made that the WallpaperType.All should stay such that we don't get deserialization issues when old config files get loaded.

Jay-o-Way:
and outside of mentioned file, example
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/48b45a6b53c908dc4d5d9a229224e784853f6598/AutoDarkModeLib/ComponentSettings/Base/WallpaperSwitchSettings.cs#L42-L50

krisbiradar:
@Spiritreader mentioned i need to remove `StateUpdate` , `TypeNeedsUpdate`  but what i understood was i need to remove the implementation of `WallpaperType.All` . i think i need to modify them to not use `WallpaperType.All` and not completely remove them.
Also will add comment and keep the enum around..

Spiritreader:
> think i need to modify them to not use `WallpaperType.All` and not completely remove them.

Yes that's correct, I was referring to the logic for WallpaperType.All within those functions.
Sorry for the misunderstanding

PR #1017, opened 2025-08-19
code cleanup  -  wallpapertype.all

### LINKED ISSUE #1015: Cleanup: WallpaperType.All

krisbiradar:
redo #1016 
closes #1015

### DIFF
--- a/AutoDarkModeLib/ComponentSettings/Base/WallpaperSwitchSettings.cs
+++ b/AutoDarkModeLib/ComponentSettings/Base/WallpaperSwitchSettings.cs
@@ -44,7 +44,7 @@ public enum WallpaperType
     Unknown = -1,
     Individual = 0,
     Global = 1,
-//    All = 2,
+    All = 2,
     SolidColor = 3,
     Spotlight = 4
 }


Spiritreader (approved):
I think that covers everything

ChenYiLins:
LGTM. With the launch of WinUI version, it is necessary for us to make some necessary amendments and improvements to the back end.

krisbiradar:
@Jay-o-Way changed it.

Spiritreader:
> There's more in other files, for example
> 
> https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/48b45a6b53c908dc4d5d9a229224e784853f6598/AutoDarkModeLib/ComponentSettings/Base/WallpaperSwitchSettings.cs#L42-L50
> 
> Tip
> 
> Try and search for "WallpaperType" in the whole project/solution, and then look for "All" in the files.
> 
> I think the and safest thing to do here is simply comment out that one line:
> 
> ```cs
>  public enum WallpaperType 
>  { 
>      Unknown = -1, 
>      Individual = 0, 
>      Global = 1, 
>      // All = 2, 
>      SolidColor = 3, 
>      Spotlight = 4 
>  } 
> ```

No. I explicitly want to leave this in because it will cause deserialization to fail if someone still uses the setting.
The deserializer can handle entries that do not exist, but if it can't map to the enum type it will reset the entire user configuration
This is undesirable.

Issue #259, opened 2021-09-25
Doesn't open with boot


mangoepic:
The App doesn't start after boot and I have to start it manually and click apply for it to start working after booting (I installed as Administrator and unchecked Login-Task)

Any ideas?

OS: Windows 21H1

Spiritreader:
We're currently in the process of developing a next major version (version X/v10)
It will release on Oct 5th, as such no new fixed for v3.1 will be pushed, as it will be replaced.

We offer bleeding edge builds to our Telegram insider group, if you're up for that.
The auto start behavior there should be fixed, and the log in task is being investigated currently.
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode#telegram-group

mangoepic:
Telegram? Interesting, usually these cults are on Discord or Slack :P

Spiritreader:
Overlord @Armin2208 is a telegram fanboi, as such we reside there!

Spiritreader:
Version X is out, please download, try again and close the issue if it works

mangoepic:
@Spiritreader Awesome! I will check it out.

Spiritreader:
On the time page, you might have to put it to disabled, and then back to the setting you had before to make it work.

mangoepic:
Hello again! Okay, the App still didn't start with boot. But what improved this time is that it automatically switched theme after I started the app and I didn't have to click "Apply" anymore! And there is a tray Icon. I really like the new features btw. Accent colour is great 🤩

I have Login-Task unchecked right now.
![image](https://user-images.githubusercontent.com/71931749/136168000-a40b1259-36da-4c3b-a53c-48e5f21f61c2.png)

If I try checking it I get the error
![image](https://user-images.githubusercontent.com/71931749/136168259-6d4f6394-f17f-4b91-8309-faa3d3d0c784.png)

But after starting it in Administrator mode it didn't give me an error.

I will try restarting my PC and see if it starts the App.

Spiritreader:
Can you attach the service.log (make sure you remove all geographic coordinates before)?
The interface tells you there is an issue, but only the service.log contains the actual error messages

mangoepic:
```
2021-10-05 15:50:16 | Info | Program.Main: commit hash: 9faf1f3 
2021-10-05 15:51:03 | Info | ZeroMQServer.Start: socket bound to port: 61849 
[...]

PR #1020, opened 2025-08-24
fix registry key being deleted when disabled via task manager

Spiritreader:
## Problem

When a registry key is set to disabled via the auto start tab in Windows Task Manager, in the current implementation, a signal is sent to the backend to immediately remove the registry key.
This operation is a willful user choice, and as such we should not interfere. See https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/issues/259#issuecomment-938840502

Windows stores the auto start entry and the enabled/disabled state in two separate keys, Auto Dark Mode only handles the creation and deletion of the primary entry. As such, you can never re-enable auto start without going into the Windows registry, 

## Solution

There are different disable states for our auto start mechanism:

1. Logon Task Disabled (not relevant for this PR)
2. Regkey not present
3. **Regkey present but set to disabled**

For case 3, we are already able to show the "Disabled" state on our Mode tab. We handle the "regkey present, but disabled" as a separate case and don't signal the backend.

## What this PR does

Prevent backend signaling when the regkey is in "present and disabled" state.

### DIFF
--- a/AutoDarkModeApp/Strings/de/Resources.resw
+++ b/AutoDarkModeApp/Strings/de/Resources.resw
@@ -905,4 +905,7 @@ Aktuell installierte Version: {0}, neue Version: {1}</value>
   <data name="Path" xml:space="preserve">
     <value>Path</value>
   </data>
+  <data name="StartWithWindowsFailed_Content" xml:space="preserve">
+    <value>Auto start has been disabled via the Task Manager "Startup Apps" tab. If you would like Auto Dark Mode to start with Windows again, please re-enable it there.</value>
+  </data>
 </root>
--- a/AutoDarkModeApp/Strings/el-gr/Resources.resw
+++ b/AutoDarkModeApp/Strings/el-gr/Resources.resw
@@ -905,4 +905,7 @@
   <data name="Path" xml:space="preserve">
     <value>Path</value>
   </data>
+  <data name="StartWithWindowsFailed_Content" xml:space="preserve">
+    <value>Auto start has been disabled via the Task Manager "Startup Apps" tab. If you would like Auto Dark Mode to start with Windows again, please re-enable it there.</value>
+  </data>
 </root>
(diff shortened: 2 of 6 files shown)


ChenYiLins:
Now it is true that Disabled can be displayed correctly in the case of "Regkey exists but is disabled".

But I began to think about another problem: suppose the user disabled the automatic start of ADM in the task manager at one time. When you want to enable it later, you will find that it can't be enabled anyway. Is this behavior regarded as a Bug when users forget that they have manually disabled it in the task manager? Is it necessary for us to make necessary prompts (such as a Dialog)?
(This option is only an optional addition.)

Spiritreader:
> Is it necessary for us to make necessary prompts (such as a Dialog)?

I think that addition could be good yes. The behavior is not a bug itself, but I agree that people might think of it as one.

We could either have a red warning text, or a dialog prompt similar to enabling the "Force DWM" options in the settings. What do you think?

ChenYiLins:
> I think that addition could be good yes. The behavior is not a bug itself, but I agree that people might think of it as one.
> 
> We could either have a red warning text, or a dialog prompt similar to enabling the "Force DWM" options in the settings. What do you think?

My suggestion is to add a dialog. And click the confirm button in the dialog, it can help users to enable (Maybe, or do nothing. Just as a reminder).

Spiritreader:
> > I think that addition could be good yes. The behavior is not a bug itself, but I agree that people might think of it as one.
> > We could either have a red warning text, or a dialog prompt similar to enabling the "Force DWM" options in the settings. What do you think?
> 
> My suggestion is to add a dialog. And click the confirm button in the dialog, it can help users to enable (Maybe, or do nothing. Just as a reminder).

Alright dialog it is then!
I think it would make sense to implement the dialog on the toggle button. As in: "if it has been disabled in the registry and user tries to re-enable it within ADM, we tell the user they should enable the entry in task manager instead".

Does that make sense to you?

ChenYiLins:
> Alright dialog it is then! I think it would make sense to implement the dialog on the toggle button. As in: "if it has been disabled in the registry and user tries to re-enable it within ADM, we tell the user they should enable the entry in task manager instead".
> 
> Does that make sense to you?

My idea is basically the same! φ(*￣0￣)

ChenYiLins (approved):

Spiritreader:
Alright, will merge this once we merge #1024 into this branch

PR #1021, opened 2025-08-25
Regression: Initialization function of ColorizationPage

ChenYiLins:
### Description

- Regression can correctly get the theme color of the current system when AccentColor is blank by default.
- At the same time, when switching to automatic, the correct theme color can also be obtained and saved to the configuration file.
- Simplify and optimize the code without any influence on the original function.

### Supplement

@Spiritreader Is the previous color acquisition logic wrong?

```c#
                if (Component.LightAutoColorization || Component.DarkAutoColorization)
                {
                    Theme requestedTheme = Theme.Unknown;
                    try
                    {
                        requestedTheme = (Theme)Enum.Parse(typeof(Theme), response.Details);
                        if (Component.LightAutoColorization && requestedTheme == Theme.Light)
                        {
                            Component.LightHex = response.Message;
                            needsSave = true;

                        }
                        if (Component.DarkAutoColorization && requestedTheme == Theme.Dark)
                        {
                            Component.DarkHex = response.Message;
                            needsSave = true;
                        }
                    }
                    catch (Exception ex)
                    {
                        ErrorMessageBoxes.ShowErrorMessage(ex, Window.GetWindow(this), "PageColorization_SaveDefaultColorization_ParseRequestedTheme");
                    }

                }
```

When the system is in light color mode, if dark theme is set to automatic acquisition, the color will not change at this time because `Component.DarkAutoColorization && requestedTheme == Theme.Dark`. I guess the accent color is different because the light and dark modes may use different wallpapers respectively. Should we keep the status quo (When the mode don't match, do nothing.), or save the accent color of the current mode?

### DIFF
--- a/AutoDarkModeApp/Views/ColorizationPage.xaml.cs
+++ b/AutoDarkModeApp/Views/ColorizationPage.xaml.cs
@@ -48,17 +48,16 @@ public sealed partial class ColorizationPage : Page
         {
             ViewModel.LightModeColorPreviewBorderBackground = new SolidColorBrush(dialogContent.InternalColorPicker.Color);
             _builder.Config.ColorizationSwitch.Component.LightHex = dialogContent.InternalColorPicker.Color.ToHex();
-        }
 
-        try
-        {
-            _builder.Save();
-            if (_builder.Config.ColorizationSwitch.Enabled) _dispatcherQueue.TryEnqueue(() => RequestThemeSwitch());
-
-        }
-        catch (Exception ex)
-        {
-            await _errorService.ShowErrorMessage(ex, App.MainWindow.Content.XamlRoot, "ColorizationPage");
+            try
+            {
+                _builder.Save();
+                if (_builder.Config.ColorizationSwitch.Enabled) _dispatcherQueue.TryEnqueue(() => RequestThemeSwitch());
+            }
+            catch (Exception ex)
+            {
+                await _errorService.ShowErrorMessage(ex, App.MainWindow.Content.XamlRoot, "ColorizationPage");
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


Spiritreader:
> Should we keep the status quo (When the mode don't match, do nothing.), or save the accent color of the current mode?

Yes, I think the status quo should be kept. The reason is that if we deviated from how it currently works and you have set, let's say, dark and light theme to automatic, then the accent color for light would be written into both `LightHex` and `DarkHex`.
But the accent colors may be different if different wallpapers are used. As such, it would technically write the wrong color into the config file because both automatic options would be populated with the light theme accent color. 
This then displays the wrong accent color preview. 

> Is the previous color acquisition logic wrong?

No I don't think it's wrong.
If we wanted to populate both at once correctly, cycling through both light and dark themes would be required. I think that introduces too much lag. As such, my recommendation is to keep the logic as is!

ChenYiLins:
> No I don't think it's wrong. If we wanted to populate both at once correctly, cycling through both light and dark themes would be required. I think that introduces too much lag. As such, my recommendation is to keep the logic as is!

Yes, I think so, too. Although there are ways to obtain different modes of accent color, the process of obtaining it is cumbersome and will bring influence to users, so let's keep the status quo.

Spiritreader:
I've synchronized the PR with the upstream winui3 branch and added a theme switch request when a color is saved and the colorization feature is enabled!

https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/1021/commits/5db5c507337f4be1bf5dfd6ec3055e766b1f20a8

I'm really not sure why git did such a large rebase on this though, the changes should already be replayed on the winui3 branch

Spiritreader:
Oh I figured it out. It is because your own winui3 branch fork diverged and that caused the changes to be applied to yours.

My apologies. It should not interfere with the content of this PR in terms of what you implemented, but I'll be more careful next time.

ChenYiLins:
This PR focuses on "Regression". If there are no other problems, let's merge it.

Issue #996, opened 2025-07-13
Bug: NavigationView Header gets cut off


Armin2208:
**Describe what issue or improvement you have with the usability design**
<img width="1199" height="861" alt="Image" src="https://github.com/user-attachments/assets/d9a60fb5-be44-4ec8-9f90-aba242bddba0" />
This is how my Auto Dark Mode App opened after performing the upgrade. Obviously way too small.

**Describe how you would like to see it improved**
Implement a minimum window size.

Jay-o-Way:
We do actually have a minimum size. We can increase the values if needed. @Armin2208 what's your Windows version? .NET version? And screen resolution? Scaling?

https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/dcec8ada1c8b7371025a1c8fe14eff2f7fac96ee/AutoDarkModeApp/MainWindow.xaml.cs#L19-L20

It seems that the Header is positioned in the wrong place, with bad/negative margin/padding. Easy to reproduce here on Windows 10. But not on my Windows 11. 😵

Jay-o-Way:
https://github.com/copilot/share/ca23518c-0be0-8c55-b901-9002a47529e7

ChenYiLins:
It's strange, with 2k resolution and 100% zoom, it looks like this when the window is the smallest.

<img width="739" height="388" alt="Image" src="https://github.com/user-attachments/assets/d86f30d1-7091-4ddf-854c-41e2edd73414" />

Jay-o-Way:
Oh boy, something funky is going on here 😕

Jay-o-Way:
Okay so the size is super easy to fix, just set **min height** to 420. The fact that the navigation (menu and header) are messed up is a different issue.

Google/AI says:

>The NavigationView control in WinUI3 on Windows 10 has several known issues, including highlighting inconsistencies, potential crashes, memory leaks, and navigation-related problems. These issues have been reported in various forums and issue trackers like [GitHub](https://github.com/microsoft/microsoft-ui-xaml/issues) and [Microsoft Q&A](https://learn.microsoft.com/en-us/answers/questions).

Full answer: https://www.google.com/search?q=known+issues+with+NavigationView+on+windows+10+in+winui3

GitHub issues:

- https://github.com/microsoft/microsoft-ui-xaml/issues/10489
- https://github.com/microsoft/microsoft-ui-xaml/issues/9789

[...]

PR #1026, opened 2025-09-01
Fix: NavigationView Header gets cut off

ChenYiLins:
### Description

Because the AppWindows related API does not handle DPI. This PR solves the navigationView header gets cut off caused by this problem.

Closes: #996 

More info: https://github.com/microsoft/microsoft-ui-xaml/issues/2731

### DIFF
--- a/AutoDarkModeApp/MainWindow.xaml
+++ b/AutoDarkModeApp/MainWindow.xaml
@@ -35,6 +35,7 @@
         <NavigationView
             x:Name="NavigationViewControl"
             Grid.Row="1"
+            CompactModeThresholdWidth="0"
             IsBackButtonVisible="Collapsed"
             IsPaneToggleButtonVisible="False"
             IsSettingsVisible="True"


Jay-o-Way:
Technically, this only prevents the NavView menu to _collapse_ at certain combination of values. When I use lower values, or set my scaling to something like 150%, the issue still remains. So, unfortunately, it's a nice try, but not THE solution.
I think setting `PaneDisplayMode="LeftCompact"` is much more effective, because it prevents the collapse of the menu. But then we will loose the labels in normal view. ⚖️

image

Jay-o-Way:
WinUI Gallery has a very custom and complicated construction, so I wouldn't take that as example. In PowerToys, they use properties **CompactModeThresholdWidth** and **ExpandedModeThresholdWidth**, both at **1007**. This might also be a solution. Let me try this...

ChenYiLins:
> Technically, this only prevents the NavView menu to _collapse_ at certain combination of values. When I use lower values, or set my scaling to something like 150%, the issue still remains. So, unfortunately, it's a nice try, but not THE solution. I think setting `PaneDisplayMode="LeftCompact"` is much more effective, because it prevents the collapse of the menu. But then we will loose the labels in normal view. ⚖️

Does this have anything to do with the resolution of the display? I tried 150% scaling at 1080p and 2k resolutions, and this problem didn't happen again.

Jay-o-Way:
@ChenYiLins I got it! setting `CompactModeThresholdWidth="0` prevents the menu from collapsing, but _does_ switch between compact and expanded menus. This is the simplest and best solution!

ChenYiLins:
> @ChenYiLins I got it! setting `CompactModeThresholdWidth="0` prevents the menu from collapsing, but _does_ switch between compact and expanded menus. This is the simplest and best solution!

Just set it in Xaml? Let me do it now. In fact, I also want to add that the current values of MinWidth and Heght are not specific values, just follow the old ADM window.

Jay-o-Way:
Oh, there's one more thing. The minimum values are calculated and set on app load, so if user changes scaling _after_ opening the app, the window can shrink back to _actual_ pixels in stead of scaled size. But in practice this is a rare case.

Spiritreader (approved):
We might have to adjust the configured minimum dimensions (870 x 600) before we release. But that's just if we notice that the ADM interface is larger than the entire screen surface. This is not to be expected, and thus shouldn't prevent us from merging this now!

Issue #989, opened 2025-07-04
Feature request: Switch different filling modes between different modes


pohuing:
**Describe the enhancement or feature you'd like**
Switching between bright and dark mode should also switch the fill mode. 

Accordingly it should move down to the wallpaper section 

<img width="2012" height="1212" alt="Image" src="https://github.com/user-attachments/assets/b42559a3-50cf-45de-9ea5-7dfc4da11dd4" />

**Describe alternatives you've considered**
Converting all of my wallpapers to 16:10.

**Additional context**
Neither of my wallpapers are 16:10, which means I always either have ugly borders where I don't want them, or a blurry mess. 

What I want:
This is a 1:1 image, so I want
Fill mode: center, "Anpassen"
<img width="2880" height="1800" alt="Image" src="https://github.com/user-attachments/assets/4f1ac0b1-10c7-47d6-b6b6-03e13d7985d1" />

And this is a 16:9 screenshot, so I want the sides cut off with
Fill mode: fill, "Füllen"
<img width="2880" height="1800" alt="Image" src="https://github.com/user-attachments/assets/b942c73b-301b-45c7-bba0-8dadcacfd6af" />

ChenYiLins:
Thank you for your proposal. At present, this part of the function is not included in the underlying design, and may be updated in the future.

Spiritreader:
Cool feature suggestion. You make an absolutely valid point that different wallpapers in Dark/Light should support different filling modes.

The required changes to make this work would be:
- [ ] Modify `WallpaperSwitchSettings.Position` and split into `WallpaperSwitchSettings.PositionLight` and `WallpaperSwitchSettings.PositionDark`
- [ ] Modify the frontend to handle the separation of this setting into Light and Dark Modes
- [ ] Update the backend WallpaperSwitchThemeFile.cs and WallpaperSwitch.cs to handle the two new modes by including a Dark/Light check for the setting

@ChenYiLins, @Jay-o-Way I think we should target 11.1 for this. Getting 11.0 out the door with the UI changes and critical fixes is prio first, but it could be a nice addition for the first feature update based on v11.

Jay-o-Way:
[...]

PR #1027, opened 2025-09-01
Fix: SettingsCard header be cut off

ChenYiLins:
### Description

- Fixed that when the path is too long, the Textblock will overwrite the Header of the SettingsCard, which will not be displayed.
- Moved **Monitor** up above image setting.

### Screenshots

image

### DIFF
--- a/AutoDarkModeApp/Views/WallpaperPickerPage.xaml
+++ b/AutoDarkModeApp/Views/WallpaperPickerPage.xaml
@@ -122,9 +122,16 @@
 
                         <!--  Picture path  -->
                         <controls:SettingsCard
-                            Description="{x:Bind ViewModel.GlobalWallpaperPath, Mode=OneWay}"
+                            HorizontalContentAlignment="Stretch"
+                            ContentAlignment="Vertical"
                             Header="{helpers:ResourceString Name=Path}"
-                            Visibility="{x:Bind ViewModel.CurrentDisplayFlags, Converter={StaticResource FlagsToVisibilityConverter}, ConverterParameter=ShowImageSettings, Mode=OneWay}" />
+                            Visibility="{x:Bind ViewModel.CurrentDisplayFlags, Converter={StaticResource FlagsToVisibilityConverter}, ConverterParameter=ShowImageSettings, Mode=OneWay}">
+                            <TextBlock
+                                Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                IsTextSelectionEnabled="True"
+                                Style="{StaticResource CaptionTextBlockStyle}"
+                                Text="{x:Bind ViewModel.GlobalWallpaperPath, Mode=OneWay}" />
+                        </controls:SettingsCard>
 
                         <!--  Picture fit  -->
                         <controls:SettingsCard Header="{helpers:ResourceString Name=Position}" Visibility="{x:Bind ViewModel.CurrentDisplayFlags, Converter={StaticResource FlagsToVisibilityConverter}, ConverterParameter=ShowFillingWaySettings, Mode=OneWay}">
--- a/AutoDarkModeApp/Views/WallpaperPickerPage.xaml
+++ b/AutoDarkModeApp/Views/WallpaperPickerPage.xaml
@@ -112,7 +112,10 @@
                         <!--  Picture path  -->
                         <controls:SettingsCard Visibility="{x:Bind ViewModel.CurrentDisplayFlags, Converter={StaticResource FlagsToVisibilityConverter}, ConverterParameter=ShowImageSettings, Mode=OneWay}">
                             <controls:SettingsCard.Header>
-                                <StackPanel Orientation="Horizontal" Spacing="8">
+                                <StackPanel
+                                    Margin="0,0,-24,0"
+                                    Orientation="Horizontal"
+                                    Spacing="8">
                                     <TextBlock Text="{helpers:ResourceString Name=PathAt}" />
                                     <TextBlock
                                         Foreground="{ThemeResource TextFillColorSecondaryBrush}"
--- a/AutoDarkModeApp/Views/WallpaperPickerPage.xaml
+++ b/AutoDarkModeApp/Views/WallpaperPickerPage.xaml
@@ -100,6 +100,17 @@
                             </RadioButtons>
                         </controls:SettingsCard>
 
+                        <!--  Monitor  -->
+                        <controls:SettingsCard Header="{helpers:ResourceString Name=Monitor}" Visibility="{x:Bind ViewModel.CurrentDisplayFlags, Converter={StaticResource FlagsToVisibilityConverter}, ConverterParameter=ShowMonitorSettings, Mode=OneWay}">
+                            <controls:SettingsCard.Description>
+                                <HyperlinkButton Click="RemoveDisconnectedMonitorsHyperlinkButton_Click" Content="{helpers:ResourceString Name=RemoveDisconnected}" />
+                            </controls:SettingsCard.Description>
+                            <ComboBox
+                                x:Name="MonitorsComboBox"
+                                HorizontalAlignment="Right"
+                                SelectedItem="{x:Bind ViewModel.SelectMonitor, Mode=TwoWay}" />
+                        </controls:SettingsCard>
+
                         <!--  Picture  -->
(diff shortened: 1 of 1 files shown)


Jay-o-Way:
True, the Header will always give way to the Content, even if Wrapping is set. Is it an idea to align the text block to the right? So that it's more in line with the rest. Maybe have to switch to a Grid for this, if I remember correctly.

ChenYiLins:
> True, the Header will always give way to the Content, even if Wrapping is set. Is it an idea to align the text block to the right? So that it's more in line with the rest. Maybe have to switch to a Grid for this, if I remember correctly.

This looks a bit strange.

image

Jay-o-Way:
Yeah, it's a tricky little thing... How about simple Header + Description?
image

It looks like this in Windows Settings.
image

P.S. my personal opinion is that using a textblock to allow `IsTextSelectionEnabled="True"` is unneeded complicated. If it were me, I would stick to plain text.

Jay-o-Way:
P.S. note that I see one example in WinUI Gallery that has the exact same issue. Also note that there may be other SettingCards with this issue:
image

Jay-o-Way:
@Spiritreader now that I am working on this xaml, I see that the setting for the **image file** IS dependent on the **Monitor**, but **Position** IS NOT dependent on **Monitor**. Touching comments from #989

Jay-o-Way:
### Update
image

ChenYiLins:
@Jay-o-Way Regarding the previous design, I think I just want to keep consistency. Just like the path in the settings page. It seems a little strange to put it in description? I don't know what your opinion.

image

About path design, we can also refer to PowerToys.

image

And `IsTextSelectionEnabled=“True”`. I think it is necessary to keep the path accessible to users, instead of searching one by one in Explorer by address.

Jay-o-Way:
My personal opinion is that, for **one** string, using the description is **perfect**. It handles styling and wrapping. As soon as there are more strings that belong together, design becomes more complicated. Notice the example from Windows Settings?

And maybe it's me, but I dont see any value of IsTextSelectionEnabled. This is an **informative** thing, the user already selected the image earlier.

Jay-o-Way:
By experimenting, I see that the content **can** wrap under the header & description, but it looks like the **SettingsCardWrapThreshold** is around `0px` width? [Default](https://github.com/CommunityToolkit/Windows/blob/5bc705f2f563549914327a04a61c4322b8fc7fef/components/SettingsControls/src/SettingsCard/SettingsCard.xaml#L110) should be `476`. I think we had a border-case, where it didn't catch the wrapping correctly.

Example: using Header, Description, Content=TextBlock (no properties) ☑️
SettingsCard Wrap Threshold

But when I have no Description and a (styled) TextBlock with a different string, wrapping simply fails. ❌
SettingsCard Wrap Fail

@niels9001 can you explain and help?

ChenYiLins:
> And maybe it's me, but I dont see any value of IsTextSelectionEnabled. This is an **informative** thing, the user already selected the image earlier.

Why I think it is necessary. Let me give you an example:

In WinUI Gallery. As an easy-to-write content, the version number is still selectable. In the face of such interactive or duplicated content, we should enable selectivity.

image

So is the path. Take my previous screenshot as an example. Although I chose the files in the path, this path is not easy to access. What can be selectively provided is a convenience.

niels9001:
@Jay-o-Way @ChenYiLins Yes, you can overwrite that threshold value by using lightweight styling. See example in the WCT Gallery app (SettingsCard sample): https://github.com/CommunityToolkit/Windows/blob/main/components%2FSettingsControls%2Fsamples%2FSettingsCardSample.xaml#L45

Jay-o-Way:
Thanks @niels9001 that's what I thought. The problem is that it looks like this Settings card is misbehaving. We are not changing that property anywhere. That's why I was hoping you could see where the issue comes from.

niels9001:
@Jay-o-Way make sure TextWrapping is enabled on those Desceiption/Content textblocks.

Another thing to try:

Set the header of the Card, and set the path as the content. Then set the ContentAlignment to Vertical and HorizontalContentAlignment to Stretch. That should give you the same effect as Header/Description.

Spiritreader:
> @Spiritreader now that I am working on this xaml, I see that the setting for the **image file** IS dependent on the **Monitor**, but **Position** IS NOT dependent on **Monitor**. Touching comments from #989

@Jay-o-Way Yes, that is because windows cannot set a filling mode per monitor. It's all or nothing.

Take a look at the API. SetWallpaper is per monitor, SetPosition is a global flag.
```c#
  [ComImport, Guid("B92B56A9-8B55-4E14-9A89-0199BBB6F93B"), InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
  public interface IDesktopWallpaper
  {
      void SetWallpaper([MarshalAs(UnmanagedType.LPWStr)] string monitorID, [MarshalAs(UnmanagedType.LPWStr)] string wallpaper);
      [return: MarshalAs(UnmanagedType.LPWStr)]
      string GetWallpaper([MarshalAs(UnmanagedType.LPWStr)] string monitorID);

      /// <summary>
      /// Gets the monitor device path.
      /// </summary>
      /// <param name="monitorIndex">Index of the monitor device in the monitor device list.</param>
      /// <returns></returns>
      [return: MarshalAs(UnmanagedType.LPWStr)]
      string GetMonitorDevicePathAt(uint monitorIndex);
      /// <summary>
      /// Gets number of monitor device paths.
      /// </summary>
      /// <returns></returns>
      [return: MarshalAs(UnmanagedType.U4)]
      uint GetMonitorDevicePathCount();

      [return: MarshalAs(UnmanagedType.Struct)]
      Rect GetMonitorRECT([MarshalAs(UnmanagedType.LPWStr)] string monitorID);

      int SetBackgroundColor([MarshalAs(UnmanagedType.U4)] uint color);
      [return: MarshalAs(UnmanagedType.U4)]
      uint GetBackgroundColor();

      void SetPosition([MarshalAs(UnmanagedType.I4)] WallpaperPosition position);
      [return: MarshalAs(UnmanagedType.I4)]
      WallpaperPosition GetPosition();

      void SetSlideshow(IntPtr items);
      IntPtr GetSlideshow();

      void SetSlideshowOptions(DesktopSlideshowDirection options, uint slideshowTick);
      [PreserveSig]
      uint GetSlideshowOptions(out DesktopSlideshowDirection options, out uint slideshowTick);

      uint AdvanceSlideshow([MarshalAs(UnmanagedType.LPWStr)] string monitorID, [MarshalAs(UnmanagedType.I4)] DesktopSlideshowDirection direction);

      DesktopSlideshowDirection GetStatus();

      bool Enable([MarshalAs(UnmanagedType.Bool)] bool enable);
  }
```

Same with the picture position in the theme file
```
[Control Panel\Desktop]
Wallpaper=%USERPROFILE%\Pictures\Wallhaven\wallhaven-x1xlmv.jpg
Pattern=
MultimonBackgrounds=1
WindowsSpotlight=0
PicturePosition=4
Wallpaper1=%USERPROFILE%\Pictures\Wallhaven\wallhaven-x1xlmv.jpg
Wallpaper2=%USERPROFILE%\Pictures\Wallhaven\wallhaven-530433.jpg
Wallpaper3=%USERPROFILE%\Pictures\Wallhaven\wallhaven-545158.png
Wallpaper4=%USERPROFILE%\Pictures\Wallhaven\wallhaven-545158.png
```

This limitation cannot be lifted on our end

Jay-o-Way:
@ChenYiLins @niels9001 Wait. I get it now. 💡 The reason for the "ugly" behavior is that **there is no minimum width for **PART_HeaderPresenter** and **SettingsCardWrapThreshold** is never reached because the minimum width of the window!** Therefore the _header_ will always give way to the _content_, even if you would set wrapping on a textblock. Note that **PART_ContentPresenter** has no **TextWrapping** property.

See https://github.com/CommunityToolkit/Windows/issues/727

ChenYiLins:
> @ChenYiLins @niels9001 Wait. I get it now. 💡 The reason for the "ugly" behavior is that **there is no minimum width for PART_HeaderPresenter and SettingsCardWrapThreshold is never reached because the minimum width of the window!** Therefore the _header_ will always give way to the _content_, even if you would set wrapping on a textblock. Note that **PART_ContentPresenter** has no **TextWrapping** property.

In other words, this behavior will happen because of SettingsCard itself. In fact, my suggestion is to add TextBlock to the Header instead of using the Header directly.

Jay-o-Way (approved):
This is one work-around that actually works. Thanks!

Issue #901, opened 2024-12-18
Window Snap background color does not match current theme


NotUNperson:
### Description

很抱歉，由于我的英语很差所以只能使用中文进行反馈。

在切换模式后，窗口拖放到屏幕顶端后出现在窗口背后的遮罩颜色的问题。正常的遮罩如下图所示，windows中透明为关闭状态，下简称“窗口背景色”
![PixPin_2024-12-19_01-19-53](https://github.com/user-attachments/assets/e179ace5-e89c-4ca2-90d6-89baa8fe8469)

在完全重启系统，不进行模式的切换的时候，颜色与windows的颜色设置相同，深色对应黑色，浅色对应白色。

通过Windows的设置发现切换颜色并不会修改“窗口背景色”，但是重启会修改，这可能是Windows的一个bug，**但是**AutoDarkMode在我继续尝试解决这个问题的时候也发现了一个bug。

我在设置中打开了“在切换主题时始终刷新DWM”这一选项，在切换模式的时候，”窗口背景色“总是与当前的颜色相反，也就是说：在深色模式下是白色，而浅色模式下是深色。

---

~~其实在自己折腾的时候是上文所说的倒着的顺序发现的，不过为了方便描述，调整了一下顺序，导致看起来有些无厘头。~~

可能我的标题表述也容易让人存在误解，还请见谅，我不知道应该如何简短描述这一问题

### Expected Behavior

由于windows自身的问题，在不开启DWM刷新的条件下，“窗口背景色”无法被正常切换，这不应该由你们承担后果。但是刷新DWM在打开的状态下，“窗口背景色”需要与当前颜色相匹配，也就是**浅色与白色对应**，**深色与黑色对应**。

### Log Data

[AutoDarkMode.zip](https://github.com/user-attachments/files/18186956/AutoDarkMode.zip)
[AutoDarkMode-aftervidoe.zip](https://github.com/user-attachments/files/18187244/AutoDarkMode-aftervidoe.zip)
aftervidoe是指录制下面的视频之后的日志

### Commit Hash, Version and Windows Build

- Commit: `2dd26a1`
- Service/App: `10.4.2.25`
- Updater: `3.1.6`
- Shell: `1.3.3.0`
- .Net: `8.0.0`
- Windows: `26100.2605`
- Arch: `X64`

### Screenshots / Videos

出现的问题：
[...]

PR #1030, opened 2025-09-04
Feat: HWND broadcast dwm refresh

Spiritreader:
Fixes DWM refreshes by utilizing HWND broadcast messages to propagate `ImmersiveColorSet` change to all Windows.

First I have to explain that DWM refresh probably isn't / wasn't the right term for this. But in our scope it meant different stages of window refreshes that we had control over by manipulating the theme file.

This also solves the lag that was introduced by our attempts to update the window themes with the old method.
It allows us to remove
- Theme color patching (previously known as "standard" DWM refresh)
- Swapping accent color back and forth on switch, responsible for extra lag (previously known as "full" DWM refresh)
- Complex checks on what kind of refresh is needed to keep the lag down. The logic is still kept, but simplified

Now that we know about these flags which showed up in @Jaylyn-Barbee's theme switch contribution to PoweTtoys, we can also pull this into ADM: 
https://github.com/microsoft/PowerToys/blob/jay/LightSwitch/src/modules/LightSwitch/LightSwitchService/ThemeHelper.cpp

We are now replicating this behavior by calling the same API with the same flags.

Functionally, from my limited testing, it appears to solve all issues with missing updates for UI components when switching between light and dark mode. So desyncs like #901 should no longer happen, and the taskbar / explorer  should no longer be stuck in the wrong mode. 
We managed to get these occurrences down a lot with our hacks, and Windows updates improved the behavior as well, but this should be the final missing piece.

Since PowerToys will also have a theme switch in the future, maybe a more open dialogue can be had about what those flags do. ADM at the moment still has quite a bit richer of a featureset, so the question is if there are more broadcast flags that could benefit the theme switching ecosystem as a whole.

For example, does setting a theme with CThemeManager also perform these broadcasts? Maybe to some extent, otherwise it wouldn't work. If so, why does it not always work as well as manually calling them?

## Frontend/UI Changes

- Removed DWM refresh settings from the settings, because it is no longer required.

### DIFF
--- a/AutoDarkModeApp/AutoDarkModeApp.csproj
+++ b/AutoDarkModeApp/AutoDarkModeApp.csproj
@@ -32,11 +32,21 @@
     <OutputPath>..\bin\Debug\adm-app\ui\</OutputPath>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)'=='Release'">
+  <PropertyGroup Condition="'$(Configuration)'=='Release|AnyCPU'">
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
-    <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
     <OutputPath>..\bin\Release\adm-app\ui\</OutputPath>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)'=='Release|x64'">
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <OutputPath>..\bin\Release\adm-app\ui\</OutputPath>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)'=='Release|ARM64'">
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <OutputPath>..\bin\Release\adm-app\ui\</OutputPath>
+  </PropertyGroup>
+
 
   <ItemGroup>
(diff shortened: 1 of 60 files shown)


Spiritreader:
@ChenYiLins @Jay-o-Way please verify that I removed everything in the frontend correctly and didn't miss anything

https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/1030/files#diff-b082e1e5f0ea697712126ba466a25d42b79a865d76cbad67f6bb01f069c132bd

https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/1030/files#diff-2c33ffae6a46f8cbcc1db3e0e7ae4d778122fc976998564efc451d07ce60ac8e

ChenYiLins:
> @ChenYiLins @Jay-o-Way please verify that I removed everything in the frontend correctly and didn't miss anything
> 
> https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/1030/files#diff-b082e1e5f0ea697712126ba466a25d42b79a865d76cbad67f6bb01f069c132bd
> 
> https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/1030/files#diff-2c33ffae6a46f8cbcc1db3e0e7ae4d778122fc976998564efc451d07ce60ac8e

It seems that all the old contents have been cleared.

But don't forget, if we don't need this localized string (`AlwaysRefreshDwm`), we need to delete it later.

Spiritreader:
> But don't forget, if we don't need this localized string (AlwaysRefreshDwm), we need to delete it later.

@ChenYiLins good point, I'll make an issue for cleaning up localized strings!

Please go ahead and approve the changes if you're happy for now, then i can merge the PR in!

Issue #999, opened 2025-07-17
Task: Details on System Areas page


Jay-o-Way:
<img width="482" height="273" alt="Image" src="https://github.com/user-attachments/assets/4eb22ec7-27a6-46d8-8db5-96508ca3ffe8" />

1. Why is this checkbox (`AdaptiveTaskbarAccentCheckBox`) always checked? It's functionless. Shouldn't this be a Header + Dropdown?
2. If this checkbox (`DWMPrevalenceSwitcCheckBox`) is **unchecked**, windows should have **no** borders. But they do, because the result is connected to the Radiobuttons only.

~Also, do these name and context match?~

ChenYiLins:
Regarding the first part, I think you are right. At first, I designed it just to make everything look completely unified, but it seems that there is indeed a logical loophole.

The second part, this should be a small defect from the back end. There is no immediate response to the cancellation of the border, which will only take effect when the Hook is triggered next time.

In fact, this Name is more in line with the configuration name in Builder.

Jay-o-Way:
Does it make sense to change the second setting to a dropdown too? Something like [None, Light, Dark, Both ] ?

ChenYiLins:
> Does it make sense to change the second setting to a dropdown too? Something like [None, Light, Dark, Both ] ?

This is definitely a good idea if the back end can be implemented smoothly.

Jay-o-Way:
Question: are "Adapt to system" and "Disabled" not the same thing? Asking as a user...

Idea: I want to use ToggleSwitches, to match Windows Settings. I **think** it is possible, but that requires an even bigger refactoring of the entire logic in the background. Need to discuss this further. Also note that a setting like "Show accent color on window titlebar and borders" is independent from Windows theme. Concept:

<img width="677" height="482" alt="Image" src="https://github.com/user-attachments/assets/2255e85d-9898-4ac6-9890-9cc405fb48c5" />

We also should respect this thing, but I think we already do that by hiding it?: 

<img width="352" height="63" alt="Image" src="https://github.com/user-attachments/assets/1962f122-205c-4548-8f09-146bb73c9d9b" />

Also, we should (re)write code that looks like the examples on MS Learn.

Example: does this setting apply to light or dark theme? is the accent color applied when ADM switches theme, the user switches it via Windows Settings, or both?

[...]

PR #1036, opened 2025-09-06
Feature: Decouple Taskbar Accent Color from System Component

### LINKED ISSUE #823: Taskbar Accent Color for Both Modes
**Describe the enhancement or feature you'd like**

Currently, we can only use the taskbar accent color either in light or dark mode.
I would like to have it enabled for both by adding a third option "Both" or similar.
Should be a simple thing, but before creating a PR myself, I'd rather collect some opinions here first.

**Describe alternatives you've considered**

Currently I switch manually. My primary motivation is to have a white font on the taskbar at all times while still having a light/dark mode in general. My wallpaper is quite dark, that's why :)

**Additional context**

image

Spiritreader:
Previously, the taskbar color was tied to the system theme due to a limitation in Windows that required turning off / on the accent color in a specific order.

Since we can now instruct windows to reload their theme via HWND broadcast, this is no longer required.
As such, we can now decouple the taskbar accent color switching into its own component.

This provides greater flexibility, including but not limited to better handling for #999.

The new options are:

```c#
public bool TaskbarColorSwitch { get; set; }
public Theme TaskbarColorDuring { get; set; } = Theme.Light;
```
and work in the same way as the accent color for title bars and window borders. 

Users can have the following states:
- Accent color for the taskbar active during light mode (TaskbarColorSwitch true and TaskbarColorDuring set to Theme.Light)
- Accent color for the taskbar active during dark mode (TaskbarColorSwitch true and TaskbarColorDuring set to Theme.Dark)
- Dsiabled (TaskbarColorSwitch false)
which might be turned into toggles instead of the radio button approach.

Accent color for the taskbar during light theme is not permitted by Windows, but actually works fine. Therefore we support this mode now. The backend will automatically resolve this state when the feature is disabled while the accent color is set for a light theme, such that it can be switched again in the Windows control panel.

### DIFF
--- a/AutoDarkModeSvc/SwitchComponents/Base/SystemSwitch.cs
+++ b/AutoDarkModeSvc/SwitchComponents/Base/SystemSwitch.cs
@@ -17,6 +17,8 @@
 using System;
 using System.Threading.Tasks;
 using AutoDarkModeLib;
+using AutoDarkModeLib.ComponentSettings.Base;
+using AutoDarkModeSvc.Events;
 using AutoDarkModeSvc.Handlers;
 
 namespace AutoDarkModeSvc.SwitchComponents.Base;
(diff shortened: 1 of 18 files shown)


AutoDarkModeSvc/SwitchComponents/Base/SystemSwitch.cs
```diff
@@ -17,17 +17,126 @@
...
+            Logger.Error(ex, "couldn't initialize system apps theme state");
+        }
+    }
+
     public override bool ThemeHandlerCompatibility { get; } = false;
-    protected override async Task SwitchSystemTheme(Theme newTheme)
+
+    protected override bool ComponentNeedsUpdate(SwitchEventArgs e)
+    {
+        if (Settings.Component.Mode == Mode.AccentOnly)
```
Jay-o-Way, line 59: This looks like a nice situation for a _switch_

Spiritreader: For me this falls under the category of backend refactors, which are not a target for 11 yet, and not the purpose of this PR
I'm trying to keep changes minimal and this is a legacy component.

So for the most part if an existing component needs to be moved around only logic changes are performed for now.

AutoDarkModeSvc/SwitchComponents/Base/TaskbarColorSwitch.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+        catch (Exception ex)
+        {
+            Logger.Error(ex, "couldn't retrieve DWM prevalence state: ");
+        }
+    }
+
+    protected override bool ComponentNeedsUpdate(SwitchEventArgs e)
+    {
+        if (e.Theme == Theme.Dark)
```
Jay-o-Way, line 37: This looks like a nice situation for a _switch_

Spiritreader: Same story as https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/1036#discussion_r2327304637

AutoDarkModeSvc/SwitchComponents/Base/TaskbarColorSwitch.cs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    protected override void HandleSwitch(SwitchEventArgs e)
+    {
+        if (e.Theme == Theme.Dark)
+        {
+            if (Settings.Component.TaskbarColorDuring == Theme.Dark)
+            {
+                RegistryHandler.SetTaskbarColorPrevalence(1);
+                currentTaskbarColorActive = true;
```
Jay-o-Way, line 69: Nothing wrong with this, but logically, this is the same as `currentTaskbarColorActive = (Settings.Component.TaskbarColorDuring == Theme.Dark)`; and SetTaskbarColorPrevalence() could also use a ternary. Same result though.

AutoDarkModeSvc/SwitchComponents/Base/SystemSwitch.cs
Jay-o-Way: Alrighty

AutoDarkModeSvc/SwitchComponents/Base/TaskbarColorSwitch.cs
Spiritreader: True, if you wanted to minimize it, the registry handler methods that set 1 and 0 would also accept a boolean, then you could do sth like

```
var settingIsLight = Settings.Component.TaskbarColorDuring == Theme.Light;
RegistryHandler.SetTaskbarColorPrevalence(settingIsLight);
currentTaskbarColorActive = settingIsLight;
```

But to me, the two "true"'s are semantically different.
One true is in the context of "is the current taskbar color active". The other one is "does the requested setting equal to the Light theme".
So in terms of (at least in my personal case) code readability: "currentTaskbarColorActive is set to does the requested setting equal to light theme" reads wrong.

Also what happens if we have more than 2 cases, writing a one liner can potentially look cluttered. Then we have different code styles in different locations.

I think things like that would be grounds for discussion in #1038 in terms of achieving consistency.

VitoItalianGamer:
It's possible for the title bars right?

ChenYiLins:
> and work in the same way as the accent color for title bars and window borders.

@VitoItalianGamer Sure.

> and work in the same way as the accent color for title bars and window borders.

PR #1049, opened 2025-09-09
Auto Dark Mode Version 11 with WinUI3

Spiritreader:
Opening this already so we can keep track of the changes.

I will most likely do a squash merge for this.


Jay-o-Way:
> I will most likely do a squash merge for this.

Smart choice 😉

PR #1052, opened 2025-09-10
Create settings card for accent modes

Spiritreader:
## Description

Far from done yet, but it at least moves the accent color settings away from the system modes, which is a start.
Still not sure how to handle the checkbox / radiobutton combo.

I think maybe turning the checkbox into a toggle would be okay, and then we move the "Choose when the accent color should be applied" description as a header for the "Light Theme" and "Dark Theme" radio buttons?

I don't know enough about winui3 to make this happen in a Microsoft-canonical way, so I'm gonna merge this in and then you can be free to modify as you see fit.

## Image

image

### DIFF
--- a/AutoDarkModeApp/Strings/en-us/Resources.resw
+++ b/AutoDarkModeApp/Strings/en-us/Resources.resw
@@ -126,9 +126,6 @@
   <data name="AccentColor" xml:space="preserve">
     <value>Accent color</value>
   </data>
-  <data name="AccentColorForTaskbar" xml:space="preserve">
-    <value>Accent Color for Taskbar</value>
-  </data>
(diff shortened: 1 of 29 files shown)


AutoDarkModeApp/Strings/en-us/Resources.resw
```diff
@@ -973,4 +967,16 @@ Currently installed version: {0}, new version: {1}</value>
   <data name="IgnoreColor" xml:space="preserve">
     <value>Ignore color</value>
   </data>
+  <data name="AccentSurfaces" xml:space="preserve">
+    <value>Accent Surfaces</value>
```
Jay-o-Way, line 971: Let's try and keep sentence-casing :)

PR #1053, opened 2025-09-10
UI brush up

Jay-o-Way:
Small changes


Jay-o-Way:
Ahh yes, it can be so simple, hahaha. Thanks again Chen!

Issue #1054, opened 2025-09-11
Bug: Time Page Selection options lag by one click


Spiritreader:
## Problem

Whenever a different radio button option is selected on the Time Page for the Switch setting, the extra options (configuring latitude longitude, enabling the light and dark time pickers) seem to not be synchronized / don't properly apply to the selected item.

I can't seem to get the proper option to load unless I select a different option, then leave the time page and reload it.
I briefly looked in the code and I think the visibility options maybe don't get properly applied on change of the radio button item, or not at the right time?

## Example

https://github.com/user-attachments/assets/c27f7242-18e0-4f13-a19e-3ebd0bb4a27c

Issue #1056, opened 2025-09-12
Bug: Changing offset values does not update UI


Spiritreader:
## Problem

When changing the offset values, the Light/Dark times are not updated unless the page is reloaded. Analogous to #1055, event listeners seem to be missing to correctly update the time fields on offset for coordinate and location service mode

Issue #1055, opened 2025-09-12
Bug: Time Page Geographic Coordinates Apply Button does not update UI


Spiritreader:
## Problem

When "From sunset to sunrise (geographic coordinates) is selected, changing the coordinates and hitting apply, the updated values are correctly written to the config file, but the UI is not updated.

If you move away from the time page and come back, it is correctly loaded, but real-time updates are not reflected

## Screenshot/GIF

![Image](https://github.com/user-attachments/assets/3f69b3ae-0b2b-4600-8753-957248c2464b)

PR #1058, opened 2025-09-14
Fix: Configuration watcher isn't working properly

ChenYiLins:
### Description

The reason for this Bug is b2bd150.

The ConfigWatcher property creates a new instance each time it is accessed. The event was added to a different FileSystemWatcher instance.

Closes: #1054 #1055 #1056

### DIFF
--- a/AutoDarkModeApp/Utils/Handlers/StateUpdateHandler.cs
+++ b/AutoDarkModeApp/Utils/Handlers/StateUpdateHandler.cs
@@ -12,80 +12,113 @@ namespace AutoDarkModeApp.Utils.Handlers;
 
 internal static class StateUpdateHandler
 {
+    private static readonly Lock _lock = new();
+    private static FileSystemWatcher? _configWatcher;
+    private static FileSystemWatcher? _scriptConfigWatcher;
+    private static readonly System.Timers.Timer _postponeRefreshTimer;
+
     private static readonly List<FileSystemEventHandler> _delegatesConfigWatcher = [];
     private static readonly List<FileSystemEventHandler> _delegatesScriptConfigWatcher = [];
     private static readonly List<ElapsedEventHandler> _delegatesTimer = [];
 
-    private static readonly DispatcherQueueTimer _debounceTimer;
+    private static readonly DispatcherQueueTimer? _debounceTimer;
     private static Action? _debounceAction;
 
     public static SecurityIdentifier SID => WindowsIdentity.GetCurrent().User!;
 
     static StateUpdateHandler()
     {
-        PostponeRefreshTimer.Interval = 2000;
+        _postponeRefreshTimer = new System.Timers.Timer(2000);
 
         try
         {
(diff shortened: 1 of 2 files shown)


Spiritreader:
Nice! Thanks for figuring it out!

Update: Nvm I'm blind

~~I have a question: The change that I made turned config watchers into nullable types, preventing them from becoming active when there is no config file present. However they weren't singletons before either.~~

~~This is what it looked like before the commit.~~

```
    private static FileSystemWatcher ConfigWatcher { get; } =
        new()
        {
            Path = AdmConfigBuilder.ConfigDir,
            Filter = Path.GetFileName(AdmConfigBuilder.ConfigFilePath),
            NotifyFilter = NotifyFilters.LastWrite,
        };

    private static FileSystemWatcher ScriptConfigWatcher { get; } =
        new()
        {
            Path = AdmConfigBuilder.ConfigDir,
            Filter = Path.GetFileName(AdmConfigBuilder.ScriptConfigPath),
            NotifyFilter = NotifyFilters.LastWrite,
        };
```

~~Why does turning it into a nullable type change its behavior?~~

```
 private static FileSystemWatcher? ScriptConfigWatcher
 {
     get
     {
         if (!File.Exists(AdmConfigBuilder.ScriptConfigPath))
         {
             return null;
         }

         return new FileSystemWatcher
         {
             Path = AdmConfigBuilder.ConfigDir,
             Filter = Path.GetFileName(AdmConfigBuilder.ScriptConfigPath),
             NotifyFilter = NotifyFilters.LastWrite,
         };
     }
 }
```

~~The getter returns a new FileSystemWatcher in exactly the same way as before.~~

Spiritreader:
Thanks for figuring it out!

I missed that the config watcher now turned into a getter property call instead of being assigned to the static varaible after i introduced the nullable type!

Spiritreader (approved):

Jay-o-Way:
Looks complicated to me. How do you know all this? 😅

Spiritreader:
> Looks complicated to me. How do you know all this? 😅

It's not that complicated, I basically just committed an object-oriented programming oopsie.
When I modified the getter to return null if no config file is present, I effectively changed the *Property*'s behavior on calling get, instead of preventing the creation of the instantiated object like I intended to.

```c#
static Prop MyProp { get; } = new();
```
instantiates a new `MyProp` object and assigns it to `Prop`.

However
```c#
static Prop MyProp { get { return new(); }
```
returns a new instance of `MyProp` whenever you attempt to read, or call get of `Prop`, which is obviously not what we want because we expect the static reference to be stable (every get call returns the *same* instance of the MyProp class).

Textbook oversight on my end that @ChenYiLins then caught and applied the singleton pattern to it 😆.
I haven't worked with C# in a while and got the purpose of getters mixed up. 
My day job involves a lot of python programming, I have no excuse hahaha.

I actually like this about C#, because it makes you think of how you want to implement something as it gives you lots of options. But sometimes it also leads to unintended behavior.

PR #1059, opened 2025-09-14
Regroup System Area Configuration Items Pt2

### LINKED ISSUE #999: Task: Details on System Areas page

Spiritreader:
Regroups the accent color settings into their own settings panels, and uses toggles wherever possible.

image

closes #999

### DIFF
--- a/AutoDarkModeApp/Views/SystemAreasPage.xaml
+++ b/AutoDarkModeApp/Views/SystemAreasPage.xaml
@@ -46,8 +46,7 @@
                                            Description="{helpers:ResourceString Name=AccentSurfacesChoose}"
                                            HeaderIcon="{ui:FontIcon Glyph=&#xEB41;}">
 
-                    <ToggleSwitch IsOn="{x:Bind ViewModel.IsTaskbarColorSwitch, Mode=TwoWay}" />
-
+                    <ToggleSwitch IsOn="{x:Bind ViewModel.IsDWMPrevalenceSwitch, Mode=TwoWay}" />
 
                     <controls:SettingsExpander.Items>
                         <controls:SettingsCard>
--- a/AutoDarkModeApp/Views/SystemAreasPage.xaml
+++ b/AutoDarkModeApp/Views/SystemAreasPage.xaml
@@ -45,7 +45,7 @@
                 <controls:SettingsExpander
                     Description="{helpers:ResourceString Name=AccentSurfacesChoose}"
                     Header="{helpers:ResourceString Name=AccentApplyTitlebarAndBorders}"
-                    HeaderIcon="{ui:FontIcon Glyph=&#xEB41;}">
+                    HeaderIcon="{ui:FontIcon Glyph=&#xE737;}">
 
                     <ToggleSwitch IsOn="{x:Bind ViewModel.IsDWMPrevalenceSwitch, Mode=TwoWay}" />
 
--- a/AutoDarkModeApp/Views/SystemAreasPage.xaml
+++ b/AutoDarkModeApp/Views/SystemAreasPage.xaml
@@ -50,15 +50,11 @@
                     <ToggleSwitch IsOn="{x:Bind ViewModel.IsDWMPrevalenceSwitch, Mode=TwoWay}" />
 
                     <controls:SettingsExpander.Items>
-                        <controls:SettingsCard>
-                            <controls:SettingsCard.Header>
-                                <StackPanel Orientation="Vertical">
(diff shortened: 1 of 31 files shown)


ChenYiLins:
Should we change RadioButton to ComboBox? This looks more like the style in Win11 settings.

AutoDarkModeApp/Views/SystemAreasPage.xaml
```diff
@@ -40,24 +40,19 @@
                     </ComboBox>
                 </controls:SettingsCard>
 
-                <!--  Accent Mode  -->
-                <controls:SettingsExpander Header="{helpers:ResourceString Name=AccentSurfaces}"
+
+                <!--  Accent Mode for title bars and window borders  -->
+                <controls:SettingsExpander Header="{helpers:ResourceString Name=AccentApplyTitlebarAndBorders}"
                                            Description="{helpers:ResourceString Name=AccentSurfacesChoose}"
```
Jay-o-Way:
```suggestion
```
I think this description is way too obvious now.

Jay-o-Way:
The radiobuttons also mean it is impossible to choose "both"? Two check-boxes gives 4 possible combinations. Don't need the Toggle switch then, which means the Settings Expander ~and Card can be merged~ can be removed.

Jay-o-Way:
Also, let's move "Windows mode" above "App mode" (like in Windows settings)

Jay-o-Way:
Larger idea: I'm thinking the words "Surface areas" are more appropriate?

Jay-o-Way:
@Armin2208 I know it looks pretty to have, but there's no shame in not having an icon for a specific setting. Just saying.
image

AutoDarkModeApp/Views/SystemAreasPage.xaml
```diff
@@ -40,24 +40,19 @@
                     </ComboBox>
                 </controls:SettingsCard>
 
-                <!--  Accent Mode  -->
-                <controls:SettingsExpander Header="{helpers:ResourceString Name=AccentSurfaces}"
+
+                <!--  Accent Mode for title bars and window borders  -->
+                <controls:SettingsExpander Header="{helpers:ResourceString Name=AccentApplyTitlebarAndBorders}"
```
Jay-o-Way:
```suggestion
                <controls:SettingsExpander Header="{helpers:ResourceString Name=AccentApplyTitlebarAndBorders}" IsExpanded="True"
```

(or see comment above)

Spiritreader:
> The radiobuttons also mean it is impossible to choose "both"? Two check-boxes gives 4 possible combinations. Don't need the Toggle switch then, which means the Settings Expander ~and Card can be merged~ can be removed.

You're not supposed to choose both. Choosing both is a No-Op in Auto Dark Mode and should be done via the Windows Settings. We're not switching anything if it's always on.

We are not trying to recreate the functionality of the Windows settings panel.

Spiritreader:
> @Armin2208 I know it looks pretty to have, but there's no shame in not having an icon for a specific setting. Just saying.

All top level settings entries have icons. Only entries within an expander don't.

Spiritreader:
> Should we change RadioButton to ComboBox? This looks more like the style in Win11 settings.

I've checked what Windows does for single choice menus, and they use radiobuttons.

image

ChenYiLins:
> > Should we change RadioButton to ComboBox? This looks more like the style in Win11 settings.
> 
> I've checked what Windows does for single choice menus, and they use radiobuttons.

Actually, I think we should use ComboBox according to the pictures. Haha. WinUI Gallery mentioned "Use a combobox when you need to conserve on-screen space and when users select only one option at a time. A combobox shows only the currently selected item". RadioButton is suitable for displaying options more intuitively or when options need some description. Light and Dark mode are both very simple options.

(But we can also use RadioButton, because this piece is in SettingsExpander. What's your opinion?)

Jay-o-Way:
@niels9001 do you have advice?

niels9001:
I'd just stick to an `SettingsExpander` with a single `SettingsCard` that uses a `ComboBox` with light/dark?

Armin2208:
This icon looks pretty ass on Windows 10
grafik

Armin2208 (approved):

PR #1068, opened 2025-09-18
Feature: Permit Spotlight in Frontend for 22H2

ChenYiLins:
### Description

Enables spotlight in the wallpaper pick page for win10 builds that have the option available.


Spiritreader (changes requested):
~~The check must follow the same logic as the backend, because technically Windows builds between Win10_22H2_Spotlight and Win11_22H2_Spotlight do not support spotlight, and this permits outdated win 11 builds to enable spotlight without actually having that functionality~~

Nvm I'm blind 😆
The github diff tool continues to confuse me

Spiritreader (approved):

PR #1071, opened 2025-09-18
Code quality: Performance improvement of TimePage

ChenYiLins:
### Decription

Now this is a completely innocuous update, which only solves the problem that the page is overloaded because of ConfigWatcher. The caching scheme is still under study...

To facilitate the review, I will briefly describe every commit.

## Code quality: Performance improvement

- Adjust the order of function blocks
- Add fields to skip ConfigWatcher updating UI, and update UI when it changes (a little faster than ConfigWatcher)

### DIFF
--- a/AutoDarkModeApp/ViewModels/TimeViewModel.cs
+++ b/AutoDarkModeApp/ViewModels/TimeViewModel.cs
@@ -19,7 +19,9 @@ public partial class TimeViewModel : ObservableRecipient
     private readonly IErrorService _errorService;
     private readonly IGeolocatorService _geolocatorService;
     private readonly Microsoft.UI.Dispatching.DispatcherQueueTimer _debounceTimer;
+
     private bool _isInitializing;
+    private bool _skipConfigUpdate;
 
     public enum TimeSourceMode
     {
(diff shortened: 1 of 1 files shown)


Jay-o-Way:
Need some help?

ChenYiLins:
> Need some help?

Ah, thank you very much!! The reason why this PR is shelved is that I want to improve the backend (Although it hasn't started yet, haha(oﾟvﾟ)ノ), so this PR should probably be postponed to Version11.1.

Jay-o-Way:
Let's also add #1078 and #1079 to this.

Spiritreader:
> > Need some help?
> 
> Ah, thank you very much!! The reason why this PR is shelved is that I want to improve the backend (Although it hasn't started yet, haha(oﾟvﾟ)ノ), so this PR should probably be postponed to Version11.1.

Functional changes to the backend are quite a bit of time away on the roadmap. 11.1 feels to early in that regard.
Every backend change needs to be heavily scrutinized and there should never be more than one active backend PR, which means modernization of code syntax and refactoring will take quite a bit more time.

For exsample: A lot of the functional decisions in the backend are made to work around a specific bug or undesired behavior by Windows (including loading and saving config files).

Spiritreader:

Update: nvm figured it out!

PR #1074, opened 2025-09-22
Code quality: Language settings

Jay-o-Way:
In this PR there are changes to the settings for app language.

- Added missing languages (Greek, Korean, Swedish, Arabic and Hebrew) in the UI 😲
- It now uses a short code (like `en` or `zh-Hans`) to store the setting.
  - renamed "Language" to "SelectedLanguageCode" in code and settings file **#BreakingChange**
- Created a Helper **GetDefaultLanguageAsync** that checks for missing or invalid data (from _LocalSettings_ file) and returns a fallback when needed.
- Fallback language when setting is not found:
  1. Top listed language in Windows Settings, if available (first regional variant, then neutral variant)
  2. Fallback to English
- **SettingsViewModel** uses an `ObservableCollection<LanguageOption>` to combine pretty names and short codes
- The **InfoBar** is not open when SelectedLanguage == current app language
- _App.xaml.cs_ --> `admConfigBuilder.Config.Tunable.UICulture = LanguageHelper.SelectedLanguageCode` before svc activation

### Screenshot

image


Jay-o-Way:
@ChenYiLins What is the difference is between ResourceExtensions.cs and ResourceHelpers.cs?

Spiritreader:
>  remove LanguageTranscoding, specifically in SettingsViewModel

```c#
 _builder.Config.Tunable.UICulture = Localization.LanguageTranscoding(Language!);`
```
Updating the AdmConfig value sets the language for the backend only. The frontend uses a separate configuration found in `%localappdata%`. At least that's how it's been with the WPF implementation. I know that at least some settings are still stored there.

> If/when this step is okay, i want to try and add instant-refresh on change.

Don't forget to still send a restart command on the backend, because that one does not support on-the-fly language switching for the tray yet.

ChenYiLins:
> @ChenYiLins What is the difference is between ResourceExtensions.cs and ResourceHelpers.cs?

Extensions are used in code, and Helpers are used in xaml.

Spiritreader:
Also: I think we need to re-set jumplists on a locale change.

ChenYiLins:
> * [ ]  If/when this step is okay, i want to try and add instant-refresh on change.

No, it's impossible. Whether we use x:Uid or not, we can't switch languages without restarting the application when using Unpackage mode.

Jay-o-Way:
> > * [ ]  If/when this step is okay, i want to try and add instant-refresh on change.
> 
> No, it's impossible. Whether we use x:Uid or not, we can't switch languages without restarting the application when using Unpackage mode.

I'm still investigating and learning, but it may be possible by reloading the window content. But indeed, then the tray menu (and who knows what more) must be reloaded too.
Starting to think it would be better to create a new branch for that. I definitely can't do that level on my own.

Spiritreader:
I think we best make a "cost vs benefit" analysis:

**How much time would it take to realize this, and how important is it?**

The reason why we never bothered with live reload is that it never mattered enough for actually using ADM and was too complicated to implement.
If we come to the same conclusion here, we should not proceed.

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
```diff
@@ -197,15 +199,43 @@ private void LoadSettings()
         {
             _isInitializing = true;
 
-            var languageText = await _localSettingsService.ReadSettingAsync<string>("Language");
-            if (languageText != null)
+            var language = await _localSettingsService.ReadSettingAsync<string>("SelectedLanguageCode");
+            if (!string.IsNullOrEmpty(language))
```
Jay-o-Way, line 203: This needs more testing

AutoDarkModeApp/App.xaml.cs
```diff
@@ -128,11 +128,10 @@ protected async override void OnLaunched(LaunchActivatedEventArgs args)
     private static async Task SetApplicationLanguageAsync()
     {
         var localSettings = App.GetService<ILocalSettingsService>();
-        var language = await localSettings.ReadSettingAsync<string>("Language");
-        if (language != null)
+        var language = await localSettings.ReadSettingAsync<string>("SelectedLanguageCode");
+        if (!string.IsNullOrEmpty(language))
```
Jay-o-Way, line 132: Needs more safety

Jay-o-Way (commented):
More checks for bad data from storage. Maybe turn the if-then code into a function?

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
```diff
@@ -197,15 +199,43 @@ private void LoadSettings()
...
+                    {
+                        SelectedLanguageCode = neutralLanguage;
+                    }
+                    else
+                    {
+                        SelectedLanguageCode = "en";
+                        //or default to <DefaultLanguage/>
+                    }
+                }
+                await _localSettingsService.SaveSettingAsync("SelectedLanguageCode", SelectedLanguageCode);
```
Jay-o-Way: @Spiritreader @ChenYiLins this whole block of code uses some nice checks to find the best language. I want to use a _something_ for this, so that I can reuse that safety feature in other places. For example in GeolocatorService.cs: **_langcode** uses `GetCultureInfo(language)`

Can you help me on my way?

Jay-o-Way: Note that `Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride` and `CultureInfo.GetCultureInfo(language).TwoLetterISOLanguageName.ToUpperInvariant()` will freak out if the data (from the settings file) is not a valid lang-code.

Spiritreader: > I want to use a something for this

what do you mean by a something?

You have a bunch of code that checks a couple of variables and sets a couple of values.
The local settings services looks like a dependency you can inject.
I'd assume that the logic for the check itself is not tied to any particular object instance.
So I'd create a static helper class that takes the service as dependency and returns the selected language cocde as well as the top language, which you can then use to assign the values wherever you need this functionality.

Spiritreader: > checks to find the best language

Can you explain what you mean by "best language?

> example in GeolocatorService.cs: 

The geolocator service should be based on the user configured language, even if it differs from the preferred system language of the OS.
If the application configures itself on first start to pick the preferred language, the GeoLocator service pulls the language preference from the ADM app settings. Not sure why we would be needing a preferred lang check there. Or am I missing sth?

Jay-o-Way: > what do you mean by a something?

A method, a function, a helper... 🤷‍♂️ I am learning things on-the-go here 😅

> Can you explain what you mean by "best language?

What I mean is more a safety check in general. Think of use cases:

- First run, there is no settings file or this setting is not in the file. → There should be a **SaveSettingAsync** as soon as possible; not waiting for Settings page to be loaded (that is: in _App.xaml.cs_ - which is currently missing)
- The saved data is a valid code, but not a part of **SupportedCultures**. → Fallback
- The saved data is invalid. → Try and set "test" as the SelectedLanguageCode in the file. → **PrimaryLanguageOverride** and `CultureInfo.GetCultureInfo(language).TwoLetterISOLanguageName` will crash.

Jay-o-Way: @ChenYiLins @Spiritreader have a look at the new **GetDefaultLanguageAsync** in the latest commit. It is used in _App.xaml.cs_ and in _GeolocatorService.cs_ and it checks the data from the settings file and corrects any mistakes (like missing or invalid string, see previous comment). At the time of writing this, it lives in **SettingsViewModel**, but it should be moved to a different place...

ChenYiLins: > So I'd create a static helper class that takes the service as dependency and returns the selected language cocde as well as the top language, which you can then use to assign the values wherever you need this functionality.

That's a good idea.

Jay-o-Way:
@Spiritreader @ChenYiLins please take another look. I think it's pretty good now. JumpList and tray menu all change after the restart.

Jay-o-Way:
Not sure if I should apply this?
image

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
```diff
@@ -202,16 +204,7 @@ private void LoadSettings()
...
-                Language = languageText.Replace("\"", "");
-            }
-            else
-            {
-                Language = "English (English)";
-                await _localSettingsService.SaveSettingAsync("Language", "English (English)");
-            }
+            SelectedLanguageCode = await LanguageHelper.GetDefaultLanguageAsync();
 
             _isInitializing = false;
```
Jay-o-Way: This can be cleaned up, right?

AutoDarkModeApp/Services/GeolocatorService.cs
Jay-o-Way, line 1: Can you check the namespace for this file? VS suggests it needs **AutoDarkModeApp.Services**

Jay-o-Way (commented):
@Spiritreader @ChenYiLins please take another look. I think it's pretty good now. JumpList and tray menu all change after the restart. But at first-run, menus default to English. 😐

Jay-o-Way:
https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/b093dd1b23fbd097a1efa15ce797e16595f545a4/AutoDarkModeSvc/Service.cs#L69

This defaults to English (until language is changed). How to fix?

Jay-o-Way:
> Updating the AdmConfig value sets the language for the backend only. The frontend uses a separate configuration found in `%localappdata%`. At least that's how it's been with the WPF implementation. I know that at least some settings are still stored there.

True. Which is strange, from a user perspective. A user that changes the language in the app will expect this to be applied everywhere. Yet we are storing this in different places... 
Should we be using one settings file???

Jay-o-Way:
Off-topic: how the h*ll did we forget to add languages? 🤣🤣🤣

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
ChenYiLins: Of course, the original code simply set the default language to English, and now `GetDefaultLanguageAsync` can work better.

AutoDarkModeApp/Services/GeolocatorService.cs
ChenYiLins: I don't have such a hint here.

ChenYiLins:
> True. Which is strange, from a user perspective. A user that changes the language in the app will expect this to be applied everywhere. Yet we are storing this in different places... Should we be using one settings file???
>
> Let me put it simply, from the developer's point of view, this is not surprising. Svc still uses Winform, so it shows different settings storage characteristics from WinUI (even I think this is a retrogression of WinUI, because it is really inconvenient to store settings! )

Okay I understand that. But now - if I want the same code in the SVC project, should I copy and paste it?

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
Jay-o-Way: Sorry, I meant the _dispatcherqueue_ block

ChenYiLins: Done.

ChenYiLins:
## Fix: Some languages cannot be selected normally

### Description

Fixed an error that some languages might not be selected normally in ComboBox during initialization.
```c#
var neutralLanguage = topLanguage.Split('-')[0]; // example: "fr"
```
Such as "pt-BR", "pt-PT". These codes will be divided into multiple segments instead of just `[0]`.

ChenYiLins:
To be honest, I still like this PR very much. Because it separates the language modules that are coupled in SettingsView and ViewModel. We can manage language additions and deletions more conveniently and call language modules everywhere. And the display of the language no longer needs hard coding by the developer (which is pleasant ヾ(≧▽≦*)o).

Jay-o-Way:
> ## Fix: Some languages cannot be selected normally
> ### Description
> Fixed an error that some languages might not be selected normally in ComboBox during initialization.
> 
> ```cs
> var neutralLanguage = topLanguage.Split('-')[0]; // example: "fr"
> ```
> 
> Such as "pt-BR", "pt-PT". These codes will be divided into multiple segments instead of just `[0]`.

@ChenYiLins What do you mean by this? **topLanguage** is always _one_ single string. The idea is that if a regional language (like `es-MX` for Mexico, which we don't have) is not available, it takes the first part `es` and checks if _that_ is available. Therefore I want to revert that last commit. More importantly, `topLanguageArray.Length > 2` can never be true, because a string like `en-US` can never be more than two parts.

> [!NOTE]
> Syntax `topLanguage.Split('-')[0]` means: "_First_ split, _then_ take element number '0' (which is the first element) from that array" Arrays are zero-based.

Or am I actually missing something?

ChenYiLins:
> @ChenYiLins What do you mean by this? **topLanguage** is always _one_ single string. The idea is that if a regional language (like `es-MX` for Mexico, which we don't have) is not available, it takes the first part `es` and checks if _that_ is available. Therefore I want to revert that last commit. More importantly, `topLanguageArray.Length > 2` can never be true, because a string like `en-US` can never be more than two parts.

Wrong. Let's see what topLanguage has gained:
```c#
            var preferredLanguages = ApplicationLanguages.Languages; // example: ["fr-FR", "en-US", "de-DE"]
            string topLanguage;
            if (preferredLanguages.Any())
            {
                topLanguage = preferredLanguages[0];
            }
            else // very unlikely, but just in case
            {
                topLanguage = CultureInfo.CurrentUICulture.Name;
            }

            System.Diagnostics.Debug.WriteLine(topLanguage);
```

Debug: `zh-Hans-CN`

`topLanguage.Split('-')[0]` will let us only get "zh". This will cause the matching to fail. Then ComboBox will not be selected. In the same way, so is "pt-BR".

At the same time, because the code of the language will not exceed two parts (for example, `ab-cd-ef` will not exist), we don't need loops to splice, just splicing [0] and [1].

Jay-o-Way:
> Debug: `zh-Hans-CN`

Ohh, my apologies, I did not know about - or think about that possibility! 🫢

ChenYiLins:
### Sceenshots

image

Jay-o-Way:
@ChenYiLins Sorry for reverting your commit before I understood why. I have created a commit that does the same thing. Also chose a different format for the strings.
image

Jay-o-Way:
### To do

```cs
CultureInfo.DefaultThreadCurrentUICulture = new CultureInfo(Builder.Config.Tunable.UICulture);
```
defaults to English at first run.

ChenYiLins:
> defaults to English at first run.

Does the previous code mean that we don't have to use English as the default language, but it is also feasible to use system preference values?

ChenYiLins:
One more thing, I guess it is the wrong opinion given by AI. In the previous code, JumpList was not initialized to English at the first run.

#### Previous code

```c#
            Debug.WriteLine("Adding jumplist items");
            Debug.WriteLine($"Current culture: {CultureInfo.CurrentCulture.Name}");
```

#### New code

```c#
            Debug.WriteLine("Adding jumplist items");
            Debug.WriteLine($"Current language: {Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride}");
```

If this is to output the language used by the current program, then it is wrong. Because this is not the language used by WinUI. As shown in the screenshot, the Debug output is Chinese, but the program actually uses English. Including JumpList. (AI can't always distinguish WinUI, WPF and Winform╰(*°▽°*)╯)

image

However, this error will not affect any practical functions, but it may bring misunderstanding to developers.

ChenYiLins:
> Okay I understand that. But now - if I want the same code in the SVC project, should I copy and paste it?

Hahaha, I don't know Jay how you edited my reply. I guess you want to add a reply, and I just saw it now. (●'◡'●)

`LanguageHelper` will not take effect in Svc because `ApplicationLanguages` is a class of WinUI. At the same time, the language switching of Svc should work normally now, and we don't need to make any changes. Because `builder.Config.Tunable.UICulture` originally uses short codes.

Jay-o-Way:
> Hahaha, I don't know Jay how you edited my reply. I guess you want to add a reply

Indeed, wrong button (Edit versus Quote). Oops.

> the language switching of Svc should work normally now, and we don't need to make any changes

Solved.

> If this is to output the language used by the current program, then it is wrong.

Good catch! Will update or remove this.

AutoDarkModeApp/App.xaml.cs
```diff
@@ -119,6 +119,9 @@ protected async override void OnLaunched(LaunchActivatedEventArgs args)
         base.OnLaunched(args);
 
         Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = await LanguageHelper.GetDefaultLanguageAsync();
+        var admConfigBuilder = AdmConfigBuilder.Instance();
```
Spiritreader, line 122: this does not respect the first run directive. You need to move this into the Activation Service and gurad it with the first run variable

https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/b093dd1b23fbd097a1efa15ce797e16595f545a4/AutoDarkModeApp/Services/ActivationService.cs#L30

Needs to be after this line but before `serviceStartIssued`

and the guard is
```
!await localSettingsService.ReadSettingAsync<bool>("NotFirstRun")
```

Spiritreader (approved):

PR #1075, opened 2025-09-23
Update .vsconfig

Jay-o-Way:
As the title

### DIFF
--- a/.vsconfig
+++ b/.vsconfig
@@ -1,13 +1,14 @@
 {
     "version": "1.0",
     "components": [
-        "Microsoft.VisualStudio.Workload.Universal", // UWP workload (needed for WinUI 3 tooling)
+        "Microsoft.NetCore.Component.Runtime.9.0", // .NET 9 runtime (preview or latest)
+        "Microsoft.NetCore.Component.SDK", // .NET SDK
+        "Microsoft.VisualStudio.Component.ManagedDesktop.Prerequisites", // Prerequisites for .NET desktop development
+        "Microsoft.VisualStudio.Component.ManagedDesktopCore", // .NET desktop development (WPF, WinForms)
+        "Microsoft.VisualStudio.Component.MSBuild", // MSBuild
+        "Microsoft.VisualStudio.Component.NuGet", // NuGet package manager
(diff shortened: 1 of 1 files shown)


Spiritreader:
> Don't take it the wrong way

None taken, starting over will result in less flexibility with changes but a clean start, whereas refactoring forces evaluation of decisions. Two different paths, both have merit. 
We're not under any time pressure, so refactoring piece by piece allows us to deliver the full feature set of the app and be release-ready while doing so.

> discovering issues and cleaning them up takes just as much time in total...

I don't fully understand if both measures take the same amount of time how that would make starting over favorable.

Jay-o-Way:
> I don't fully understand if both measures take the same amount of time how that would make starting over favorable.

That must be a personal preference.

Before merging, please do check if there's anything missing...

.vsconfig
```diff
@@ -1,16 +1,11 @@
 {
-  "version": "1.0",
-  "components": [
-    "Microsoft.Component.MSBuild",
```
ChenYiLins, line 4: Can this line be removed? This is used to build.

ChenYiLins: I checked it in the `installer` just now, and it can't be deleted because WinUI depends on it.

.vsconfig
```diff
@@ -1,16 +1,11 @@
...
-    "Microsoft.NetCore.Component.Runtime.7.0",
-    "Microsoft.NetCore.Component.SDK",
-    "Microsoft.VisualStudio.Component.ManagedDesktop.Core",
```
ChenYiLins, line 7: And this. `Svc` is built in Winform, should it not be deleted?

Jay-o-Way: thanks, good catches! _(had my head in winui/wasdk)_

.vsconfig
```diff
@@ -1,16 +1,13 @@
...
-    "Microsoft.VisualStudio.Component.ManagedDesktop.Prerequisites",
```
ChenYiLins, line 8: This seems to be what winform needs. ( •̀ ω •́ )

Jay-o-Way: done

ChenYiLins:
Is there anything left out? If not, we can merge it.

PR #1077, opened 2025-09-26
Rename color filter to reflect new changes

Armin2208:
Still working on it, isn't ready yet

### DIFF
--- a/AutoDarkModeApp/Strings/de/Resources.resw
+++ b/AutoDarkModeApp/Strings/de/Resources.resw
@@ -214,7 +214,7 @@ In einigen Fällen kann ein Piepston auftauchen, wenn die Maus während der DWM 
     <value>Farbe</value>
   </data>
   <data name="ColorFilter" xml:space="preserve">
-    <value>Farbfilter (Sendet Win + Strg + C)</value>
+    <value>Farbfilter während des Dark Modes</value>
   </data>
   <data name="ColorizationPick_Description" xml:space="preserve">
     <value>Passe die Windows Akzentfarbe nach deiner Präferenz an.</value>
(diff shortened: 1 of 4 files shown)


Armin2208:
grafik
Current state. I consider renaming "Adapt to system" to "Enable Switching" and renaming "Disabled" to "Disable switching". Maybe you have ideas for a better wording.
Also, maybe it makes sense to expand the accent color settingsbox by default

Armin2208:
Also we should add descriptions at the top of pages, just like win11 settings, to describe what the user can expect from the current page.
What is System Ares? What is Personalization?
grafik

Jay-o-Way:
@Armin2208 what would you like to add to this pr?

Armin2208:
grafik

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
```diff
@@ -512,6 +512,7 @@ partial void OnIsDwmRefreshViaColorizationChanged(bool value)
                 XamlRoot = App.MainWindow.Content.XamlRoot,
                 CloseButtonText = "Cancel".GetLocalized(),
                 PrimaryButtonText = "Confirm".GetLocalized(),
+                PrimaryButtonStyle = (Style)App.Current.Resources["AccentButtonStyle"],
```
Jay-o-Way, line 515: It would have been better to use
```cs
        DefaultButton = ContentDialogButton.Primary
```

https://learn.microsoft.com/windows/apps/design/controls/dialogs-and-flyouts/dialogs#:~:text=Use%20the%20ContentDialog.DefaultButton%20property%20to%20indicate%20the%20default%20button.%20By%20default%2C%20no%20default%20button%20is%20set.

Armin2208: That's a really nice and easy way to accomplish the same. I changed that and pushed in to winui3. 65bb3f1adfbed3a8a7845468cb17c967a9c906f3

PR #1084, opened 2025-10-05
UX: Disable controls versus hiding them

Jay-o-Way:
As title.
Disabled Controls

### DIFF
--- a/AutoDarkModeApp/Views/HotkeysPage.xaml
+++ b/AutoDarkModeApp/Views/HotkeysPage.xaml
@@ -6,12 +6,18 @@
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:helpers="using:AutoDarkModeApp.Helpers"
     xmlns:local="using:AutoDarkModeApp.Views"
+    xmlns:converters="using:CommunityToolkit.WinUI.Converters"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     mc:Ignorable="d">
+    <Page.Resources>
+        <converters:BoolNegationConverter x:Key="BoolNegationConverter" />
+    </Page.Resources>
 
     <ScrollViewer>
         <Grid x:Name="ContentArea">
             <StackPanel Style="{StaticResource PageBaseStackPanelStyle}">
+
+                <InfoBar Message="{helpers:ResourceString Name=HotkeyInfoBarMessage}" IsOpen="True" IsClosable="False" Severity="Informational"/>
 
                 <!--  Settings  -->
                 <TextBlock Style="{StaticResource SettingsSectionHeaderTextBlockStyle}" Text="{helpers:ResourceString Name=Settings}" />
@@ -48,7 +54,7 @@
                             <controls:SettingsCard
                                 Margin="0,0,0,4"
                                 Header="{x:Bind DisplayName}"
-                                IsEnabled="{Binding IsOn, ElementName=HotkeysToggleSwitch}">
+                                IsEnabled="{Binding IsOn, ElementName=HotkeysToggleSwitch, Converter={StaticResource BoolNegationConverter}}">
                                 <StackPanel Orientation="Horizontal" Spacing="8">
                                     <TextBox
                                         MinWidth="160"
--- a/AutoDarkModeApp/Strings/ar/Resources.resw
+++ b/AutoDarkModeApp/Strings/ar/Resources.resw
@@ -144,11 +144,6 @@
   <data name="AlwaysLight" xml:space="preserve">
     <value>دائماً فاتح</value>
   </data>
-  <data name="AlwaysRefreshDwmMsg_Content" xml:space="preserve">
-    <value>محاولات لتقليل حدوث ألوان واجهة المستخدم غير الصحيحة في مستكشف الملفات، شريط المهام، وقائمة ابدأ. إذا كنت تواجه هذه المشكلة بشكل متكرر أو تجدها مزعجة، حاول تمكين هذا الإعداد.
-
-قد يؤدي ذلك إلى تأخير طفيف إلى متوسط في بعض الأنظمة أثناء تغيير السمات، مشابه لما حدث في Windows 10. في حالات نادرة، قد يصدر Windows صوت تنبيه عند تحريك الماوس أثناء تحديثات DWM.</value>
-  </data>
   <data name="Apply" xml:space="preserve">
     <value>طبق</value>
   </data>
@@ -979,14 +974,19 @@
   <data name="AggressiveDwmRefresh" xml:space="preserve">
     <value>Aggressive DWM refresh</value>
   </data>
-  <data name="AggressiveDwmRefreshMsg_Title" xml:space="preserve">
-    <value>Use aggressive DWM refresh?</value>
-  </data>
   <data name="OffsetTime_Header" xml:space="preserve">
     <value>Delay switching by a number minutes</value>
   </data>
   <data name="OffsetTime_Description" xml:space="preserve">
     <value>Positive values are later, negative values are earlier</value>
+  </data>
+  <data name="AlwaysRefreshDwmMsg_Content" xml:space="preserve">
+    <value>محاولات لتقليل حدوث ألوان واجهة المستخدم غير الصحيحة في مستكشف الملفات، شريط المهام، وقائمة ابدأ. إذا كنت تواجه هذه المشكلة بشكل متكرر أو تجدها مزعجة، حاول تمكين هذا الإعداد.
+
+قد يؤدي ذلك إلى تأخير طفيف إلى متوسط في بعض الأنظمة أثناء تغيير السمات، مشابه لما حدث في Windows 10. في حالات نادرة، قد يصدر Windows صوت تنبيه عند تحريك الماوس أثناء تحديثات DWM.</value>
+  </data>
+  <data name="AggressiveDwmRefreshMsg_Title" xml:space="preserve">
+    <value>Use aggressive DWM refresh?</value>
   </data>
   <data name="DarkModeForWebbrowser" xml:space="preserve">
     <value>Dark mode for your web browser</value>
@@ -1014,5 +1014,8 @@
   </data>
   <data name="OpenWindowsSettings" xml:space="preserve">
     <value>Open Windows settings</value>
+  </data>
+  <data name="HotkeyInfoBarMessage" xml:space="preserve">
+    <value>Hotkeys can only be changed while disabled</value>
   </data>
 </root>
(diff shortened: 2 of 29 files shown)


Jay-o-Way:
@Armin2208 @Spiritreader @ChenYiLins Please review?

AutoDarkModeApp/Views/HotkeysPage.xaml
```diff
@@ -7,6 +7,9 @@
     xmlns:helpers="using:AutoDarkModeApp.Helpers"
     xmlns:local="using:AutoDarkModeApp.Views"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:viewmodels="using:AutoDarkModeApp.ViewModels"
+    x:Name="HotkeysRoot"
+    d:DataContext="{d:DesignInstance Type=viewmodels:HotkeysViewModel}"
```
ChenYiLins: Do we need them?

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
```diff
@@ -21,8 +21,9 @@ public sealed partial class HotkeysPage : Page
 
     public HotkeysPage()
     {
-        ViewModel = App.GetService<HotkeysViewModel>();
         InitializeComponent();
+        ViewModel = App.GetService<HotkeysViewModel>();
+        DataContext = ViewModel;
```
ChenYiLins: At the same time, the statements here are in the wrong order, so we should ensure that `ViewModel` is initialized before `InitializeComponent()`.

AutoDarkModeApp/Views/HotkeysPage.xaml
Jay-o-Way: Yes, these are **crucial** to bridging the scope of the **ItemsRepeater**. ViewModel doesn't work when using inside it, because it is linked to the page.

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
Jay-o-Way: Can you explain?

Jay-o-Way: image

AutoDarkModeApp/Views/HotkeysPage.xaml
ChenYiLins: I mean `d:DataContext`, winui doesn't have a designer. And even if `Binding` is used, it is to access the DataContext of the Page, instead of directly binding with `DataContext`. Nothing will happen without this line of code.

ChenYiLins: At the same time, I want to add that I think this method is over-engineered. This is an event at the UI level, so we can just operate it directly in View.

Jay-o-Way: Try to remove one or two of these lines and see what happens to the input controls for the hotkeys.

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
ChenYiLins: What the hell is AI talking about? ? ? I mean ViewModel. ┗|｀O′|┛

Jay-o-Way: > What the hell is AI talking about? ? ? I mean ViewModel. ┗|｀O′|┛

You selected line 25 and 26?
But it doesn't matter. Why do you insist that **InitializeComponent** must come after ViewModel?

Jay-o-Way:
@ChenYiLins changing the binding to the Toggle switch is inconsistent with the other pages. It should work, but I usually prefer a binding to the view model, just to be sure it reflects the actual state of the view model, not just a control.

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
ChenYiLins: Let me make it simple. The `InitializeComponent()` method parses XAML and creates UI elements. Because there is data binding in XAML, we should ensure that the data source is ready before binding occurs (that is initialization of **ViewModel**). In fact, we didn't feel this error because the code was executed too fast, but in fact this is the initialization order of the error. **DataContext** should be the same, which is why I selected lines 25-26.

I wrote a piece of code to let you feel this process intuitively:

```c#
    public HotkeysPage()
    {
        InitializeComponent();

        LoadViewModel();
    }

    private async Task LoadViewModel()
    {
        System.Diagnostics.Debug.WriteLine("Start the time");
        await Task.Delay(5000);
        System.Diagnostics.Debug.WriteLine("Time is up");
        ViewModel = App.GetService<HotkeysViewModel>();
    }
```

#### Sceenshots

Since **ViewModel** is not initialized at first, all bindings that use `x:Bind` will not take effect and will return to normal after initializing **ViewModel**.

PixPin_2025-10-13_14-10-16

ChenYiLins:
> @ChenYiLins changing the binding to the Toggle switch is inconsistent with the other pages. It should work, but I usually prefer a binding to the view model, just to be sure it reflects the actual state of the view model, not just a control.

```xaml
<ToggleSwitch x:Name="HotkeysToggleSwitch" IsOn="{x:Bind ViewModel.HotkeysEnabled, Mode=TwoWay}" />
```

I don't quite understand what you mean. Toggle's IsOn is bound to the fields in ViewModel in both directions (TwoWay), and the binding IsOn and ViewModel are actually in the same state.

Jay-o-Way:
> > @ChenYiLins changing the binding to the Toggle switch is inconsistent with the other pages. It should work, but I usually prefer a binding to the view model, just to be sure it reflects the actual state of the view model, not just a control.
> 
> ```xaml
> <ToggleSwitch x:Name="HotkeysToggleSwitch" IsOn="{x:Bind ViewModel.HotkeysEnabled, Mode=TwoWay}" />
> ```
> 
> I don't quite understand what you mean. Toggle's IsOn is bound to the fields in ViewModel in both directions (TwoWay), and the binding IsOn and ViewModel are actually in the same state.

You are using

```cs
 IsEnabled="{Binding IsOn, ElementName=HotkeysToggleSwitch}">
```

Which **should** work, but it doesn't link to the view model.  If, for any (rare) reason, there is an error between the control and saving the data, user would not see. It's a bit of a principle matter. But that is why I had the other solution: actually hook into the view model, via the page.

Armin2208:
You can only edit Hotkeys if Hotkeys are disabled. So they aren't allowed to be disabled controls if the feature is disabled.
Please test before making changes!!! This is behaviour which obviously doesn't work. Don't implement stuff without testing it.

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
Jay-o-Way: Looks like I owe you an apology. I was convinced the order was right the other way around, in part because it simply worked. My memory was based on different info. Part of it is that using `Binding` or `x:Bind` is an important difference.
🙇🏻‍♂️

ChenYiLins:
> Which **should** work, but it doesn't link to the view model. If, for any (rare) reason, there is an error between the control and saving the data, user would not see. It's a bit of a principle matter. But that is why I had the other solution: actually hook into the view model, via the page.

This is what I find strange, since the control state is bound to **ViewModel** in both directions, unless the binding fails (which will be shown in the compilation stage) or ViewModel has asynchronous logic. Otherwise, the state of this control reflects the actual state of **ViewModel**. However, your method is indeed purer in the MVVM architecture, but as I said before, it is technically complicated, and the View layer does not need to do nothing (is this a misunderstanding of MVVM hahah? ╰(*°▽°*)╯ )

ChenYiLins:
> You can only edit Hotkeys if Hotkeys are disabled. So they aren't allowed to be disabled controls if the feature is disabled. Please test before making changes!!! This is behaviour which obviously doesn't work. Don't implement stuff without testing it.

Ah, ah, ah, we are all caught in the misunderstanding of technical discussion! This is really not feasible in practice. For example, as we discussed in #702 before, you need to disable the hotkey function to edit hotkeys, which is too inconvenient.

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
ChenYiLins: Right! `Binding` is runtime parsing, and `x:Bind` and **ViewModel** are ignored in the answer given by AI.

Jay-o-Way:
> You can only edit Hotkeys if Hotkeys are disabled. So they aren't allowed to be disabled controls if the feature is disabled. Please test before making changes!!! This is behaviour which obviously doesn't work. Don't implement stuff without testing it.

Alright, Armin, I hear you. We all make mistakes.
Would a adding a converter - that inverts the Boolean value - to the binding make you happy?
Disabled Hotkeys

Armin2208:
We current version without disabling anything is okay. But I think it's better if hotkey editing is disabled while hotkeys are enabled. So I like this change. 
The information message is also nice, but it is enough to write ""Hotkeys can only be changed while disabled". Before we had this information as description for the on/off toggle. But it wasn't probably realy visible.  Maybe it is not a bad idea to put this message above the "Enable hotkey" toggle. Also the additional spacing for the information is not necessary.

Jay-o-Way:
@Armin2208 done

Spiritreader:

Jay-o-Way:
> You can only edit Hotkeys if Hotkeys are disabled. So they aren't allowed to be disabled controls if the feature is disabled. Please test before making changes!!! This is behaviour which obviously doesn't work. Don't implement stuff without testing it.

YES THEY CAN! PLEASE TAKE SOME OF YOUR OWN ADVICE! @Armin2208
Hotkeys work

Armin2208:
> YES THEY CAN! PLEASE TAKE SOME OF YOUR OWN ADVICE! @Armin2208

What happens, if:

1. You set the force dark mode hotkey to be ctrl + 1.
2. enable hotkeys
3. You actually want now that force light mode to be ctrl + 1, so you open the edit screen of force light mode hotkey and press ctrl + 1

Spiritreader:
Don't shout at our poor Armin. He's doing his best to keep things user centered

Jay-o-Way:
> Don't shout at our poor Armin. He's doing his best to keep things user centered

But, he started 😢😭

Spiritreader:
> > Don't shout at our poor Armin. He's doing his best to keep things user centered
> 
> But, he started 😢😭

It was a bit of an unfortunate formulation on his end.

The problem is if you try to move an already-set hotkey to a different function. Because the backend will be listening for that specific hotkey, and setting it in the frontend while it's already assigned will cause it to trigger when it's unwanted.

PR #1090, opened 2025-10-11
Code quality: Practical update of ColoriztionPage

ChenYiLins:
### Description

Waiting for long time, I'm back ~~~ ( •̀ ω •́ )✧

- Redesign ColoriztionPage
- Added a palette similar to the one selected in win11 settings (But as Sam said before, we have no intention of turning ADM into another software that references Win11 settings, so the palette part is just for the convenience of users, but I don't want to do a 1:1 replica exactly according to the settings, so it's just similar in function and appearance)
- Added a desktop preview indicator (When switching to Light and Dark mode, it is not a Bug that Accent color will not change. You can also refer to what [Sam said before](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/1021#issuecomment-3225756922))
- Improve HandleConfigUpdate, ConfigWatcher will no longer be triggered repeatedly (it will bring a little performance improvement, although it is very small)

### Sceenshots

image

PixPin_2025-10-11_11-27-09

### DIFF
--- a/AutoDarkModeApp/Views/ColorizationPage.xaml.cs
+++ b/AutoDarkModeApp/Views/ColorizationPage.xaml.cs
@@ -17,8 +17,8 @@ public sealed partial class ColorizationPage : Page
 
     public ColorizationPage()
     {
-        InitializeComponent();
         ViewModel = App.GetService<ColorizationViewModel>();
+        InitializeComponent();
 
         PopulateColorGrid();
     }


AutoDarkModeApp/UserControls/DesktopPreview.xaml
```diff
@@ -0,0 +1,63 @@
...
+                Margin="8"
+                Padding="12"
+                HorizontalAlignment="Right"
+                VerticalAlignment="Bottom"
+                Background="{ThemeResource AcrylicInAppFillColorDefaultBrush}"
+                CornerRadius="{StaticResource ControlCornerRadius}"
+                RequestedTheme="{x:Bind DesktopPreviewThemeMode, Mode=OneWay}">
+                <Border
+                    Height="1"
+                    Margin="0,0,0,8"
```
Jay-o-Way, line 37: Might as well use a Spacing

AutoDarkModeApp/UserControls/DesktopPreview.xaml.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        var wallPaperPath = new StringBuilder(200);
+        if (SystemParametersInfo(SPI_GETDESKWALLPAPER, 200, wallPaperPath, 0) && !string.IsNullOrEmpty(wallPaperPath.ToString()))
+        {
+            return new BitmapImage(new Uri(wallPaperPath.ToString()));
+        }
+        else if (string.IsNullOrEmpty(wallPaperPath.ToString()))
+        {
+            string? colorStr = Registry.GetValue(@"HKEY_CURRENT_USER\Control Panel\Colors", "Background", "")?.ToString();
+
```
Jay-o-Way, line 81: What does all this do?

AutoDarkModeApp/Views/ColorizationPage.xaml
```diff
@@ -16,62 +18,296 @@
...
+
                 <!--  Colorization  -->
                 <TextBlock Style="{StaticResource SettingsSectionHeaderTextBlockStyle}" Text="{helpers:ResourceString Name=ColorizationPick_Description}" />
                 <controls:SettingsCard Header="{helpers:ResourceString Name=EnableAccentColorSwitch}" HeaderIcon="{ui:FontIcon Glyph=&#xE790;}">
                     <ToggleSwitch IsOn="{x:Bind ViewModel.IsColorizationEnabled, Mode=TwoWay}" />
                 </controls:SettingsCard>
 
-                <StackPanel Spacing="{StaticResource SettingsCardSpacing}" Visibility="{x:Bind ViewModel.ThemeSettingsExpanderVisibility, Mode=OneWay}">
-                    <!--  Light  -->
+                <!--  Theme  -->
```
Jay-o-Way: Themes are something different. Beter use "mode".

AutoDarkModeApp/Views/ColorizationPage.xaml
```diff
@@ -16,62 +18,296 @@
...
+                <controls:SettingsCard
+                    Header="{helpers:ResourceString Name=PickTheme}"
+                    HeaderIcon="{ui:FontIcon Glyph=&#xEA99;}"
+                    IsEnabled="{x:Bind ViewModel.IsColorizationEnabled, Mode=OneWay}">
+                    <controls:Segmented SelectedIndex="{x:Bind ViewModel.SelectColorThemeMode, Converter={StaticResource EnumToIndexConverter}, Mode=TwoWay}">
+                        <controls:SegmentedItem
+                            Content="{helpers:ResourceString Name=Light}"
+                            Icon="{ui:FontIcon Glyph=&#xE706;}"
+                            ToolTipService.ToolTip="Light" />
```
Jay-o-Way: Hardcoded. Why needed?

AutoDarkModeApp/Views/ColorizationPage.xaml
```diff
@@ -16,62 +18,296 @@
...
-                                        CornerRadius="{StaticResource ControlCornerRadius}" />
-                                    <!--  TODO: Replace Content to Res  -->
-                                    <Button
-                                        MinWidth="120"
-                                        Click="LightModeCheckColorButton_Click"
-                                        Content="{helpers:ResourceString Name=SelectColor}" />
-                                </StackPanel>
+                            <controls:SettingsCard>
+                                <controls:SettingsCard.Header>
+                                    <StackPanel Orientation="Vertical" Spacing="6">
```
Jay-o-Way:
```suggestion
                                    <StackPanel Orientation="Vertical" Spacing="8">
```

AutoDarkModeApp/Views/ColorizationPage.xaml
```diff
@@ -16,62 +18,296 @@
...
-                                        Background="{x:Bind ViewModel.LightModeColorPreviewBorderBackground, Mode=OneWay}"
-                                        CornerRadius="{StaticResource ControlCornerRadius}" />
-                                    <!--  TODO: Replace Content to Res  -->
-                                    <Button
-                                        MinWidth="120"
-                                        Click="LightModeCheckColorButton_Click"
-                                        Content="{helpers:ResourceString Name=SelectColor}" />
-                                </StackPanel>
+                            <controls:SettingsCard>
+                                <controls:SettingsCard.Header>
```
Jay-o-Way: Shouldn't this be left-aligned Content?

AutoDarkModeApp/Views/ColorizationPage.xaml.cs
```diff
@@ -1,71 +1,63 @@
...
 
-    private async void LightModeCheckColorButton_Click(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
+    private void WindowsColorGridView_ItemClick(object sender, ItemClickEventArgs e)
     {
-        var dialogContent = new ColorPickerDialogContentControl();
-        dialogContent.InternalColorPicker.IsAlphaEnabled = false;
-        dialogContent.InternalColorPicker.Color = _builder.Config.ColorizationSwitch.Component.LightHex.ToColor();
-        var colorPickerDialog = new ContentDialog()
+        var rectangleItem = e.ClickedItem as Rectangle;
+        var solidColorBrush = rectangleItem?.Fill as SolidColorBrush;
```
Jay-o-Way, line 29: Could this ever be null?

AutoDarkModeApp/Views/ColorizationPage.xaml
```diff
@@ -16,62 +18,296 @@
...
                 <TextBlock Style="{StaticResource SettingsSectionHeaderTextBlockStyle}" Text="{helpers:ResourceString Name=ColorizationPick_Description}" />
                 <controls:SettingsCard Header="{helpers:ResourceString Name=EnableAccentColorSwitch}" HeaderIcon="{ui:FontIcon Glyph=&#xE790;}">
                     <ToggleSwitch IsOn="{x:Bind ViewModel.IsColorizationEnabled, Mode=TwoWay}" />
                 </controls:SettingsCard>
 
-                <StackPanel Spacing="{StaticResource SettingsCardSpacing}" Visibility="{x:Bind ViewModel.ThemeSettingsExpanderVisibility, Mode=OneWay}">
-                    <!--  Light  -->
+                <!--  Theme  -->
+                <controls:SettingsCard
+                    Header="{helpers:ResourceString Name=PickTheme}"
```
Jay-o-Way: In general: let's keep the strings nice and simple. I'm seeing a couple of "Select..." (as this header) or "CurrentlySelected..." where they don't need to be that verbose.

Jay-o-Way:
Some time ago I was thinking of simply using Community Toolkit their Color picker button control.
https://learn.microsoft.com/dotnet/communitytoolkit/windows/colorpicker/

ChenYiLins:
> Some time ago I was thinking of simply using Community Toolkit their Color picker button control. https://learn.microsoft.com/dotnet/communitytoolkit/windows/colorpicker/

Ah! ! This is a good choice. Compared with now, it is much simpler than creating every Rectangle by myself (It seems that I have done a lot of useless work before, haha).

At the same time, I went to see some source codes of ColorPicker. Localization is also a work that we don't need to do, because Microsoft provides corresponding functions: `ColorHelper.ToDisplayName()`. Some changes must be made to this PR before the review can continue.

Regarding your questions and requests for changes, I will reply later (I need to get some sleep in my time, haha).

Armin2208:
Thanks Chen for this great Pull Request, looks great.

Please beware that we will compile the final version of Auto Dark Mode 11 on Monday! So please finish all your work until tomorrow @ChenYiLins @Jay-o-Way

AutoDarkModeApp/Strings/ar/Resources.resw
Armin2208, line 1: We are not Microsoft - So i don't thing naming every Accent Color is necessary. It is much work for the translators.

AutoDarkModeApp/UserControls/DesktopPreview.xaml
ChenYiLins: The `Margin` of the last Border is not consistent, so it is not appropriate to use `Spacing`.

AutoDarkModeApp/UserControls/DesktopPreview.xaml.cs
ChenYiLins: When `SystemParametersInfo` is Null, it means that the background is solid color, so just read the registry directly.

AutoDarkModeApp/Views/ColorizationPage.xaml
ChenYiLins: Well, I don't know why, in theory. However, when Content is put in, it will not be displayed in the correct disabled state, but it is indeed disabled at this time (because you can't click anything).

image

AutoDarkModeApp/Views/ColorizationPage.xaml.cs
ChenYiLins: It is true that there may be null when using as for conversion, just to avoid unexpected situations and make the code more secure.

AutoDarkModeApp/Views/ColorizationPage.xaml
ChenYiLins: Thanks for reminding me! This is a legacy of the previous code. Let's delete it.

AutoDarkModeApp/Strings/ar/Resources.resw
ChenYiLins: Done! All localization about colors has been removed.

ChenYiLins:
> Thanks Chen for this great Pull Request, looks great.
> 
> Please beware that we will compile the final version of Auto Dark Mode 11 on Monday! So please finish all your work until tomorrow @ChenYiLins @Jay-o-Way

Now it looks like it should be finished.

AutoDarkModeApp/UserControls/DesktopPreview.xaml.cs
Jay-o-Way: And the block below here, with all the pixel things?

ChenYiLins: No, a solid background is just one color. Like this:

image

AutoDarkModeApp/UserControls/DesktopPreview.xaml.cs
```diff
@@ -0,0 +1,117 @@
...
+                using (var stream = bitmap.PixelBuffer.AsStream())
+                {
+                    stream.Write(pixels, 0, pixels.Length);
+                }
+                return bitmap;
+            }
+        }
+        else
+        {
```
Jay-o-Way, line 90: Sorry, I was actually looking at all this... What does this do? Why not just set the background color of the control?

Jay-o-Way:
And an actual crash, because of the Error dialogs
image

Jay-o-Way:
> * Added a desktop preview indicator (When switching to Light and Dark mode, it is not a Bug that Accent color will not change.

If it's a **preview**, then the selected color **should** be shown in the preview indicator

AutoDarkModeApp/Views/ColorizationPage.xaml
Jay-o-Way: > However, when Content is put in, it will not be displayed in the correct disabled state, but it is indeed disabled at this time

Another UX bug

AutoDarkModeApp/UserControls/DesktopPreview.xaml.cs
ChenYiLins: Thank you for reminding me, I really complicated it!

ChenYiLins:
> If it's a **preview**, then the selected color **should** be shown in the preview indicator

The selected color will be fed back to the indicator, but only if the system theme pattern matches `the Theme Segmented`.

ChenYiLins:
> **2 - Bug**
> 
> * Select one of the pre-set colors
> * See the selection border
> * ComboBox will set to Manual
> * Switch back to Automatic
> * See the selection border is still on the previous color
> * And there is NO preview of the automatic color

This is the way we choose to deal with it now, not a Bug. When the light and dark modes of the system don't match, we don't really get the automatic color. This is naturally not in the color selector.

> **3 - Bug**
> 
> * This allows extremely light or dark colors as accent color, which can lead to horrible contrast or readability. Windows has a built-in check for this. It will warn the user with light or dark colors, and it will simply not apply near-white or near-black colors.

Is this actually part of the design? ADM didn't choose to check the contrast a long time ago, because the color is actually chosen by the user himself. We can also add corresponding tips to tell the user later.

ChenYiLins:
> **1 - I keep getting loads and loads of this**

~Can you tell me the steps to reproduce this problem? I have never met such a mistake.~

I know! This is because `Srv` is not running, hahaha φ(*￣0￣)

Jay-o-Way:
> > **1 - I keep getting loads and loads of this**
> 
> ~Can you tell me the steps to reproduce this problem? I have never met such a mistake.~
> 
> I know! This is because `Srv` is not running, hahaha φ(*￣0￣)

Normally these are in the background. Visible in the Output window. These keep popping up and blocking the window. Something must have changed. I should be able to debug without this.

ChenYiLins:
> Normally these are in the background. Visible in the Output window. These keep popping up and blocking the window. Something must have changed. I should be able to debug without this.

```c#
var messageRaw = MessageHandler.Client.SendMessageAndGetReply(Command.GetCurrentColorization);
```

Getting the accent color requires back-end support.

Spiritreader:
> But as Sam said before, we have no intention of turning ADM into another software that references Win11 settings, so the palette part is just for the convenience of users, but I don't want to do a 1:1 replica exactly according to the settings, so it's just similar in function and appearance)

I think this is a great idea, because the replication is directly useful for controlling colors in the context of theme switcihng. Looks great!

Spiritreader:
> > Normally these are in the background. Visible in the Output window. These keep popping up and blocking the window. Something must have changed. I should be able to debug without this.
> 
> ```cs
> var messageRaw = MessageHandler.Client.SendMessageAndGetReply(Command.GetCurrentColorization);
> ```
> 
> Getting the accent color requires back-end support.

Yes!

@Jay-o-Way you generally should never test frontend stuff without the backend running. There's multiple tie-ins and you might just get undesired behavior.

Spiritreader:
@ChenYiLins Just a quick side note: Since there's a separate component for the desktoppreview now, we can also reuse this in the future on the wallpaper page!

ChenYiLins:
> @ChenYiLins Just a quick side note: Since there's a separate component for the desktoppreview now, we can also reuse this in the future on the wallpaper page!

Yes! On the one hand, I want to keep the focus of PR and more things we need to consider in WallpaperPickPage (such as different previews in Light and Dark Mode).

Spiritreader:

Update: nvm, figured it out!

PR #1091, opened 2025-10-12
Code quality: Window-related Bug fixes and saving and reading window state

### LINKED ISSUE #1081: Code quality: check window position
Scenario:

- Open app window
- Minimize app window
- Kill app process

Result: window position is saved as `-32000, -32000`. Next time user clicks on the tray icon, window opens way out of monitor bounds. Not _impossible_ to solve**, but still good to have a check in the code. Also it would be nice to move the window back into working area, if window "sticks out" on any side.

**
- Press Alt+Space
- Select "Move"
- Press a directional key
- Move the mouse

ChenYiLins:
- Fix bugs related to WinUI `TitleBar` components
- Fix the background color of caption button hovering (this is a Bug caused by `MicaBackdrop`)
- The status of the window will now be additionally saved and read (Closed #1081)

### DIFF
--- a/AutoDarkModeApp/MainWindow.xaml.cs
+++ b/AutoDarkModeApp/MainWindow.xaml.cs
@@ -1,7 +1,6 @@
 ﻿using System.Diagnostics;
 using System.Runtime.InteropServices;
 using AutoDarkModeApp.Contracts.Services;
-using Microsoft.UI;
 using Microsoft.UI.Windowing;
 using Microsoft.UI.Xaml;
 using Windows.UI;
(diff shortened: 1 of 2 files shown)


AutoDarkModeApp/MainWindow.xaml
```diff
@@ -20,18 +19,19 @@
...
             IsBackButtonEnabled="{x:Bind NavigationFrame.CanGoBack, Mode=OneWay}"
             IsBackButtonVisible="True"
             IsPaneToggleButtonVisible="True"
             PaneToggleRequested="NavViewTitleBar_PaneToggleRequested">
-            <TitleBar.IconSource>
-                <ImageIconSource ImageSource="/Assets/AutoDarkModeIcon.ico" />
-            </TitleBar.IconSource>
+            <!--  This is a workaround for https://github.com/microsoft/microsoft-ui-xaml/issues/10374, once fixed we should just be using IconSource  -->
         </TitleBar>
+        <ImageIcon
```
Jay-o-Way, line 30: In PowerToys, they use \<TitleBar.LeftHeader> which seems a little bit cleaner 🤏

Armin2208:
Thanks for taking care!

AutoDarkModeApp/MainWindow.xaml
ChenYiLins: I know this. But now it is forced, because the LeftHeader will make the TitleBar taller, but our title bar has nothing at all, and the taller title bar becomes particularly strange.

PR #1098, opened 2025-10-14
update readme also in 11.1

Armin2208:


Armin2208:
this fucking weblate translation file is weird and github UI is making weird stuff wtf

Spiritreader:
> this fucking weblate translation file is weird and github UI is making weird stuff wtf

Yes, because, as I said, if you use a translation platform it's always going to be a nightmare when changing the strings. Which is why technically all translation MUST be paused until we are 100000000000000% sure we never want to change the translation files again.

Spiritreader:
Merging into v11.1 from master causes a cyclic hell. I've manually merged your changes into v11.1 via CLI. Closing this as the changes have been applied

Issue #1083, opened 2025-10-05
Task: synchronize UX - hide versus disable controls


Jay-o-Way:
It's best practice to **disable** settings that aren't available because of a "master" on/off toggle switch. Synchronizing this behavior means changes to:

- [ ] Time
- [ ] Background
- [ ] Accent color
- [ ] Mouse pointers
- [ ] (Windows) Theme
- [ ] Hotkeys

Armin2208:
I also prefer disabling the battery powered devices section under conditions instead of hiding it, just to "promote" the possible features for mobile devices, if someone has a secondary device.

Jay-o-Way:
> I also prefer disabling the battery powered devices section under conditions instead of hiding it, just to "promote" the possible features for mobile devices, if someone has a secondary device.

That is a different situation. A battery is not a setting that can be turned on and off at will.

If the device has no battery, showing battery-related options would be:

- Confusing: Users might wonder why those settings exist or whether something is broken.
- Cluttered: Unnecessary options dilute the clarity of the UI.
- Misleading: Disabled battery settings imply the feature is temporarily unavailable — not permanently irrelevan

ChenYiLins:
I found this problem in the development of another PR a week ago. Options are not unadjustable when the function is turned off, and we should not hide them (users will have an unknown fear and think about what will happen when the ToggleSwitch is turned on). If we choose to show these options, it is a good choice to disable them when the function is turned off.

(But due to some things, I have been too busy recently, and this PR is still on hold)

Jay-o-Way:
I think this is all done now, right?

PR #1118, opened 2025-10-18
enable hotkeys controls when function is enabled

Jay-o-Way:
Closes: #702, #1083
Because hotkeys can be changed while enabled.


ChenYiLins:
Ah, it seems that #1119 will conflict with this PR.

Spiritreader:
> Ah, it seems that https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/1119 will conflict with this PR.

Yes, please coordinate if a combination of the PRs is possible.

Jay-o-Way:
Ah Yes. @ChenYiLins since these changes are not so big, can you copy them into your branch?

Spiritreader:
@ChenYiLins @Jay-o-Way Is this PR still relevant?

ChenYiLins:
> @ChenYiLins @Jay-o-Way Is this PR still relevant?

About ThemePickerPage still makes sense. But my suggestion is that we can create a new PR. ( •̀ ω •́ )✧

PR #1119, opened 2025-10-18
Code quality and Fix: Front end and back end about hotkey function

ChenYiLins:
### Description

This is a huge PR, but it doesn't actually involve any function update. The only function change is that we can edit hotkeys when the hotkey function is disabled. See it in https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/issues/702#issuecomment-3418007928. Let me briefly explain this PR:

- Change the front-end and back-end to use [`Virtual-Key`](https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes) at the same time (this is not the exclusive benefit of Windows App SDK, this comes from win32). Because the [`RegisterHotKey`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerhotkey) already uses `Virtual-Key`, it is not necessary to convert `Winform's Keys` into VK.
- The configuration file needs to be changed. In the past, each key was connected by `+`, but now we change it to **Decorative key code** and **primary key code**, which are connected by `,`.

Before:

```yaml
Hotkeys:
  Enabled: true
  ForceLight: Ctrl + D1
  ForceDark: 
```

After:

```yaml
Hotkeys:
  Enabled: true
  ForceLight: 2,49
  ForceDark: 
```

- So the first is the change of the back end **HotkeyHandler**. There is no need to convert `Keys` to VK in the `Register` function. We only need to split the configuration obtained by `builder`, and we will get the original `Modifiers` and `keys`. Just call the `RegisterHotKey` directly.

Before:

```c#
public class HotkeyInternal
{
    public int Id { get; set; }
    public string StringCode { get; set; }
    public List<Keys> Modifiers { get; set; } = new();
    public Keys Key { get; set; }
    public Action Action { get; set; }
}
```

After:

```c#
public class HotkeyInternal
{
    public int Id { get; set; }
    public string StringCode { get; set; }
    public uint Modifiers { get; set; }
    public uint KeyCode { get; set; }
    public Action Action { get; set; }
}
```

- Then there is the change about the **Service**, and this part will become extremely simple! The `WM_HOTKEY` message originally uses `Virtual-Key` Code: the lower 16 bits of `lParam` are decorative key marks, and the upper 16 bits are virtual keycode.
- Then the front end part was modified by MVVM, and the `HotkeysDataObject` was put into **Models**. Then the original way of reading keys is modified. Just convert the read `Virtual-Key`. (It's just one more step ( •̀ ω •́ )✧)

Code:

```c#
    private static string GetKeyDisplayName(VirtualKey key)
    {
        return key switch
        {
            VirtualKey.Enter => "Enter",
            VirtualKey.Escape => "Esc",
            VirtualKey.Space => "Space",
            VirtualKey.Back => "Backspace",
            VirtualKey.Delete => "Del",
            VirtualKey.PageUp => "PgUp",
            VirtualKey.PageDown => "PgDn",
            VirtualKey.CapitalLock => "CapsLock",
            (VirtualKey)188 => ",",
            (VirtualKey)190 => ".",
            (VirtualKey)191 => "/",
            (VirtualKey)187 => "=",
            (VirtualKey)189 => "-",
            (VirtualKey)219 => "[",
            (VirtualKey)221 => "]",
            (VirtualKey)220 => "\\",
            (VirtualKey)186 => ";",
            (VirtualKey)222 => "'",
            (VirtualKey)192 => "`",
            _ => key.ToString(),
        };
    }
```

### Screenshots

PixPin_2025-10-18_23-31-32

That's all about the explanation. I hope this is a good step to start updating the 11.X version.
At the same time, @Spiritreader @Armin2208 need you to check more. I'm worried that some minor mistakes will ruin the back end. Thank you very much!

### DIFF
--- a/AutoDarkModeSvc/Handlers/HotkeyHandler.cs
+++ b/AutoDarkModeSvc/Handlers/HotkeyHandler.cs
@@ -1,174 +1,190 @@
 ﻿#region copyright
-//  Copyright (C) 2022 Auto Dark Mode
-//
-//  This program is free software: you can redistribute it and/or modify
-//  it under the terms of the GNU General Public License as published by
-//  the Free Software Foundation, either version 3 of the License, or
-//  (at your option) any later version.
-//
-//  This program is distributed in the hope that it will be useful,
-//  but WITHOUT ANY WARRANTY; without even the implied warranty of
-//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-//  GNU General Public License for more details.
-//
-//  You should have received a copy of the GNU General Public License
-//  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// Copyright (C) 2025 Auto Dark Mode
+// This program is free software under GNU GPL v3.0
 #endregion
 using System;
 using System.Collections.Generic;
-using System.Globalization;
-using System.Linq;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
-using System.Windows.Forms;
 using AutoDarkModeLib;
 using AutoDarkModeLib.Configs;
 using AutoDarkModeSvc.Core;
 
 namespace AutoDarkModeSvc.Handlers;
 
-static class HotkeyHandler
+internal static class HotkeyHandler
 {
-    private static readonly NLog.Logger Logger = NLog.LogManager.GetCurrentClassLogger();
-    private static AdmConfigBuilder builder = AdmConfigBuilder.Instance();
+    private static readonly NLog.Logger _logger = NLog.LogManager.GetCurrentClassLogger();
+
     public static Service Service { get; set; }
+
     [DllImport("user32.dll")]
     private static extern bool RegisterHotKey(IntPtr hWnd, int id, uint fsModifiers, uint vk);
-    // Unregisters the hot key with Windows.
+
     [DllImport("user32.dll")]
     private static extern bool UnregisterHotKey(IntPtr hWnd, int id);
-    public enum ModifierKeys
-    {
-        Alt = 0x1,
-        Control = 0x2,
-        Shift = 0x4,
-        Win = 0x8
-    }
-    private static List<HotkeyInternal> Registered { get; } = new();
 
+    private static List<HotkeyInternal> Registered { get; } = [];
 
     public static void RegisterAllHotkeys(AdmConfigBuilder builder)
     {
-        Logger.Debug("registering hotkeys");
+        _logger.Debug("registering hotkeys");
         try
         {
             GlobalState state = GlobalState.Instance();
 
-            if (builder.Config.Hotkeys.ForceDark != null) Register(builder.Config.Hotkeys.ForceDark, () =>
+            if (builder.Config.Hotkeys.ForceDark != null)
             {
-                Logger.Info("hotkey signal received: forcing dark theme");
-                state.ForcedTheme = Theme.Dark;
-                ThemeHandler.EnforceNoMonitorUpdates(builder, state, Theme.Dark);
-                ThemeManager.UpdateTheme(new(SwitchSource.Manual, Theme.Dark));
-            });
-
-            if (builder.Config.Hotkeys.ForceLight != null) Register(builder.Config.Hotkeys.ForceLight, () =>
-            {
-                Logger.Info("hotkey signal received: forcing light theme");
-                state.ForcedTheme = Theme.Light;
-                ThemeHandler.EnforceNoMonitorUpdates(builder, state, Theme.Light);
-                ThemeManager.UpdateTheme(new(SwitchSource.Manual, Theme.Light));
-            });
-
-            if (builder.Config.Hotkeys.NoForce != null) Register(builder.Config.Hotkeys.NoForce, () =>
-            {
-                Logger.Info("hotkey signal received: stop forcing specific theme");
-                state.ForcedTheme = Theme.Unknown;
-                ThemeHandler.EnforceNoMonitorUpdates(builder, state, Theme.Light);
-                ThemeManager.RequestSwitch(new(SwitchSource.Manual));
-            });
-
-            if (builder.Config.Hotkeys.ToggleTheme != null) Register(builder.Config.Hotkeys.ToggleTheme, () =>
-            {
-                Logger.Info("hotkey signal received: toggle theme");
-                ThemeManager.SwitchThemeAutoPauseAndNotify();
-            });
-
-            if (builder.Config.Hotkeys.ToggleAutoThemeSwitch != null) Register(builder.Config.Hotkeys.ToggleAutoThemeSwitch, () =>
-            {
-                Logger.Info("hotkey signal received: toggle automatic theme switch");
-                AdmConfig old = builder.Config;
-                state.SkipConfigFileReload = true;
-                builder.Config.AutoThemeSwitchingEnabled = !builder.Config.AutoThemeSwitchingEnabled;
-                builder.Save();
-                ToastHandler.InvokeAutoSwitchToggleToast();
-            });
-
-            if (builder.Config.Hotkeys.TogglePostpone != null) Register(builder.Config.Hotkeys.TogglePostpone, () =>
-            {
-                if (builder.Config.AutoThemeSwitchingEnabled)
-                {
-                    if (state.PostponeManager.IsSkipNextSwitch)
+                Register(
+                    builder.Config.Hotkeys.ForceDark,
+                    () =>
                     {
-                        state.PostponeManager.RemoveSkipNextSwitch();
-                        ToastHandler.InvokePauseAutoSwitchToast();
-                        Task.Delay(TimeSpan.FromSeconds(2)).ContinueWith(o => ThemeManager.RequestSwitch(new(SwitchSource.Manual)));
+                        _logger.Info("hotkey signal received: forcing dark theme");
+                        state.ForcedTheme = Theme.Dark;
+                        ThemeHandler.EnforceNoMonitorUpdates(builder, state, Theme.Dark);
+                        ThemeManager.UpdateTheme(new(SwitchSource.Manual, Theme.Dark));
                     }
-                    else
+                );
+            }
+
+            if (builder.Config.Hotkeys.ForceLight != null)
+            {
+                Register(
+                    builder.Config.Hotkeys.ForceLight,
+                    () =>
                     {
-                        state.PostponeManager.AddSkipNextSwitch();
-                        ToastHandler.InvokePauseAutoSwitchToast();
+                        _logger.Info("hotkey signal received: forcing light theme");
+                        state.ForcedTheme = Theme.Light;
+                        ThemeHandler.EnforceNoMonitorUpdates(builder, state, Theme.Light);
+                        ThemeManager.UpdateTheme(new(SwitchSource.Manual, Theme.Light));
                     }
-                }
-            });
+                );
+            }
+
+            if (builder.Config.Hotkeys.NoForce != null)
+            {
+                Register(
+                    builder.Config.Hotkeys.NoForce,
+                    () =>
+                    {
+                        _logger.Info("hotkey signal received: stop forcing specific theme");
+                        state.ForcedTheme = Theme.Unknown;
+                        ThemeHandler.EnforceNoMonitorUpdates(builder, state, Theme.Light);
+                        ThemeManager.RequestSwitch(new(SwitchSource.Manual));
+                    }
+                );
+            }
+
+            if (builder.Config.Hotkeys.ToggleTheme != null)
+            {
+                Register(
+                    builder.Config.Hotkeys.ToggleTheme,
+                    () =>
+                    {
+                        _logger.Info("hotkey signal received: toggle theme");
+                        ThemeManager.SwitchThemeAutoPauseAndNotify();
+                    }
+                );
+            }
+
+            if (builder.Config.Hotkeys.ToggleAutoThemeSwitch != null)
+            {
+                Register(
+                    builder.Config.Hotkeys.ToggleAutoThemeSwitch,
+                    () =>
+                    {
+                        _logger.Info("hotkey signal received: toggle automatic theme switch");
+                        AdmConfig old = builder.Config;
+                        state.SkipConfigFileReload = true;
+                        builder.Config.AutoThemeSwitchingEnabled = !builder.Config.AutoThemeSwitchingEnabled;
+                        builder.Save();
+                        ToastHandler.InvokeAutoSwitchToggleToast();
+                    }
+                );
+            }
+
+            if (builder.Config.Hotkeys.TogglePostpone != null)
+            {
+                Register(
+                    builder.Config.Hotkeys.TogglePostpone,
+                    () =>
+                    {
+                        if (builder.Config.AutoThemeSwitchingEnabled)
+                        {
+                            if (state.PostponeManager.IsSkipNextSwitch)
+                            {
+                                state.PostponeManager.RemoveSkipNextSwitch();
+                                ToastHandler.InvokePauseAutoSwitchToast();
+                                Task.Delay(TimeSpan.FromSeconds(2)).ContinueWith(o => ThemeManager.RequestSwitch(new(SwitchSource.Manual)));
+                            }
+                            else
+                            {
+                                state.PostponeManager.AddSkipNextSwitch();
+                                ToastHandler.InvokePauseAutoSwitchToast();
+                            }
+                        }
+                    }
+                );
+            }
         }
         catch (Exception ex)
         {
-            Logger.Error(ex, "could not register hotkeys:");
+            _logger.Error(ex, "could not register hotkeys:");
         }
-
     }
 
     public static void UnregisterAllHotkeys()
     {
-        Logger.Debug("removing hotkeys");
+        _logger.Debug("removing hotkeys");
         Registered.ForEach(hk => Unregister(hk.Id));
         Registered.Clear();
     }
 
-    public static HotkeyInternal GetRegistered(List<Keys> modifiersPressed, Keys key)
+    public static HotkeyInternal GetRegistered(uint modifiers, uint keyCode)
     {
-        return Registered.Find(hk => Enumerable.SequenceEqual(hk.Modifiers.OrderBy(e => e), modifiersPressed.OrderBy(e => e)) && hk.Key == key);
+        return Registered.Find(hk => hk.Modifiers == modifiers && hk.KeyCode == keyCode);
     }
 
     private static void Register(string hotkeyString, Action action)
     {
         if (Service == null)
         {
-            Logger.Error("service not instantiated while trying to register hotkey");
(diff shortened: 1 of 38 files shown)


Spiritreader:
hey @ChenYiLins!

Thanks for the detailed explanation.

Could you please adjust your formatter / linter to match the current format?
image

It makes it very hard to review this PR if there is mixed syntax and functional changes.


There's also another concern. Since the backend stores the keycodes in the config file, it will invalidate all current hotkey configurations.

Also
```
Hotkeys:
  Enabled: true
  ForceLight: 2,49
  ForceDark: 
```

There is no way to tell what 2,49 is, whereas Ctrl + D1 is more humanly readable.
I would like to keep the yaml configuration file verbose to a point where you can understand what's going on.

AutoDarkModeSvc/Service.cs
```diff
@@ -472,24 +472,11 @@ protected override void WndProc(ref Message m)
...
-            List<Keys> modifiersPressed = new();
-            if ((modifiers & (int)HotkeyHandler.ModifierKeys.Alt) != 0) modifiersPressed.Add(Keys.Alt);
-            if ((modifiers & (int)HotkeyHandler.ModifierKeys.Shift) != 0) modifiersPressed.Add(Keys.Shift);
-            if ((modifiers & (int)HotkeyHandler.ModifierKeys.Control) != 0) modifiersPressed.Add(Keys.Control);
-            if ((modifiers & (int)HotkeyHandler.ModifierKeys.Win) != 0) modifiersPressed.Add(Keys.LWin);
-
-            var match = HotkeyHandler.GetRegistered(modifiersPressed, key);
-            if (match == null)
-            {
-                if (modifiersPressed.Contains(Keys.LWin))
```
Spiritreader, line 486: Would your change correctly handle the case that you can press both window keys at once?

AutoDarkModeSvc/Handlers/HotkeyHandler.cs
```diff
@@ -1,216 +1,204 @@
...
+        {
+            _logger.Error(ex, $"Failed to register hotkey: {hotkeyString}");
+        }
     }
 
     private static void Unregister(int id)
     {
         if (Service == null)
         {
-            Logger.Error("service not instantiated while trying to unregister hotkey");
```
Spiritreader, line 171: Please don't do syntax changes or variable renaming in a functional PR. 
This creates clutter when attempting to review for changes. This is extremely important for all PRs going forward touching the backend.

AutoDarkModeSvc/Handlers/HotkeyHandler.cs
```diff
@@ -1,216 +1,204 @@
...
-        keys.ForEach((k) =>
+        try
         {
-            if (IsModifier(k)) mappedHotkey.Modifiers.Add(k);
-            else mappedHotkey.Key = k;
-        });
+            var parts = hotkeyString.Split(',');
+            if (parts.Length != 2 || !uint.TryParse(parts[0], out uint modifiers) || !uint.TryParse(parts[1], out uint keyCode))
+            {
+                _logger.Error($"Invalid hotkey format: {hotkeyString}");
```
Spiritreader: The hotkey string will map to the numbered keycodes instead of verbose key names. This makes it difficult to read / parse / troubleshoot when looking at the log file.

Spiritreader (changes requested):

AutoDarkModeSvc/Service.cs
ChenYiLins: I'm a little confused. shouldn't the right Win key still be displayed as Win key? (just like right Ctrl and left Ctrl). We shouldn't need to capture two key presses at the same time, just as we won't capture holding down the left and right Ctrl at the same time. And I find [it](https://superuser.com/questions/1138399/what-is-the-purpose-of-the-right-windows-key).

AutoDarkModeSvc/Handlers/HotkeyHandler.cs
ChenYiLins: Ok, I recovered in the new commit.

AutoDarkModeSvc/Handlers/HotkeyHandler.cs
ChenYiLins: Done!

AutoDarkModeSvc/Service.cs
Spiritreader: > I'm a little confused. shouldn't the right Win key still be displayed as Win key? (just like right Ctrl and left Ctrl). We shouldn't need to capture two key presses at the same time, just as we won't capture holding down the left and right Ctrl at the same time. And I find [it](https://superuser.com/questions/1138399/what-is-the-purpose-of-the-right-windows-key).

The problem I had with the hotkey handling was that configured hotkeys could either be set as LWin or RWin, and they weren't able to be used interchangeably. If that works fine we can disregard my comment.

Spiritreader:
if I understand correctly, what you're going for with the HotkeyConverter is to have a mapping from the hotkeys that arrive through the WndProc function and remap them to something that is better readable and more standardized.

Is there any chance we might encounter some edge cases on where the hotkey conversion fails? And in worst case scenario, we still get the actual raw keycode numbers, correct?

Because if we take control of the mapping, we'd have the advantage that we could document what keycodes are available in the wiki!

ChenYiLins:
@Spiritreader I made some changes according to your reply.

> Could you please adjust your formatter / linter to match the current format?

Ah, sorry. I'm using **CSharpier**. I can hardly control these behaviors, but nowI have stopped using it to make it easier to check.

> It makes it very hard to review this PR if there is mixed syntax and functional changes.
> 
> There's also another concern. Since the backend stores the keycodes in the config file, it will invalidate all current hotkey configurations.

Now I have added **HotkeyConverter** to Lib's **Helper** (most of the code is based on my previous commit). To put it simply, I chose to use the value of VK Codes before, and now it is written in the configuration file too, but it is more readable: 

```c#
return key.ToString();
```

Indeed, in fact, **HotkeyStringConverter** plays the role of translation.

> There is no way to tell what 2,49 is, whereas Ctrl + D1 is more humanly readable. I would like to keep the yaml configuration file verbose to a point where you can understand what's going on.

So now the configuration file has almost returned to the previous readable state, and even I think it is more friendly to the front-end and back-end, because as I said before, `RegisterHotKey` originally uses `Virtual-Key`.

> Is there any chance we might encounter some edge cases on where the hotkey conversion fails? And in worst case scenario, we still get the actual raw keycode numbers, correct?

In fact, your guess should exist. If we decompile **VirtualKey**, we can see that in C#, it is missing the key values of 173-194 and 219-254. Although some values can't be seen in our common keyboard types at all, `SpecialKeyMap` is to make up for this defect, that is to say, if we make up all these defects, edge cases will no longer exist.

More info: 
- https://github.com/microsoft/microsoft-ui-xaml/issues/2159
- https://stackoverflow.com/questions/32098242/how-to-handle-missing-values-from-virtualkey-enumeration-in-keyroutedeventargs-e

ChenYiLins:
## Code quality: Converter to custom name

### Screenshots

PixPin_2025-10-19_20-48-11

PixPin_2025-10-19_20-44-08

AutoDarkModeSvc/Service.cs
ChenYiLins: Just now, I set a right Win key with the keyboard mapping tool (because I can't find a keyboard with a right Win key), but the good news is that regardless of the left Win key or the right Win key, hotkey processing will treat them as Win keys.

ChenYiLins:
Then I need to add that I tried to support the old version of hotkeys, but I found that this could not be completely done.
On my keyboard, `;` The key was `Ome1` in the previous version of the capture, because each keyboard manufacturer has different definitions of keys, and we can't adapt such keys. The only feasible thing is to give the user a prompt after detecting such keys, so that the user can re-enter hot keys.

Armin2208:
What happens, if:

1. You set the force dark mode hotkey to be ctrl + 1.
2. enable hotkeys
3. You actually want now that force light mode to be ctrl + 1, so you open the edit screen of force light mode hotkey and press ctrl + 1

Spiritreader:
> Then I need to add that I tried to support the old version of hotkeys, but I found that this could not be completely done. On my keyboard, `;` The key was `Ome1` in the previous version of the capture, because each keyboard manufacturer has different definitions of keys, and we can't adapt such keys. The only feasible thing is to give the user a prompt after detecting such keys, so that the user can re-enter hot keys.

Do manufacturer hotkeys always map to the current set of virtual keys?

> In fact, your guess should exist. If we decompile VirtualKey, we can see that in C#, it is missing the key values of 173-194 and 219-254. Although some values can't be seen in our common keyboard types at all, 

If these are suuuper uncommon, I think it's okay to not process them and just write the key number into the file. Because in that case I think it also would've been difficult to get the names for those hotkeys with the old mapping. So I don't think we need to handle all of them. What do you think @ChenYiLins?

Spiritreader:
> What happens, if:
> 
>     1. You set the force dark mode hotkey to be ctrl + 1.
> 
>     2. enable hotkeys
> 
>     3. You actually want now that force light mode to be ctrl + 1, so you open the edit screen of force light mode hotkey and press ctrl + 1

@ChenYiLins If this is an issue, we could temporarily disable the hotkey functionality by setting the config option in the yaml file to false and sving it when opening the edit dialog. Then afterwards, we could re-enable it when the user has finished configuring their hotkey.

Armin2208:
> @ChenYiLins If this is an issue, we could temporarily disable the hotkey functionality by setting the config option in the yaml file to false and sving it when opening the edit dialog. Then afterwards, we could re-enable it when the user has finished configuring their hotkey.

I think this is an epic solution! But we should also beware that if the user closes the window (X or altf4) while the edit dialog is opened, it should also recover the switch hotkey ability. Because if someone clicks on the edit button, closes the window with the X, he kinda loses his hotkey abilities.

Jay-o-Way:
I can add another (strange) use case, for the _current_ master branch:

- When I press (e.g.) <kbd>Win</kbd> + <kbd>NumPad5</kbd>; ADM hotkey input says "Win NumPad5" ✔️
- When I press <kbd>Shift</kbd> + {a NumPad number}, the key-function that gets returned is what it would be if **NumLock** would be off. So
  - Pressing <kbd>Win</kbd> + <kbd>Shift</kbd> + <kbd>NumPad1</kbd> shows "Win End" ✖️
  - Pressing <kbd>Win</kbd> + <kbd>Shift</kbd> + <kbd>NumPad2</kbd> shows "Win + Down" which doesn't get picked up by the ADM input window at all. It minimizes the current window.

In this PR:
- When I press **Win** and **Shift**, it shows "Shift Win"
- When I press **NumPad1** too, it shows "Win End" and it switches back as soon as I let go.

Using a **MS wireless keyboard 3050**, in case it matters.

Spiritreader:
> > @ChenYiLins If this is an issue, we could temporarily disable the hotkey functionality by setting the config option in the yaml file to false and sving it when opening the edit dialog. Then afterwards, we could re-enable it when the user has finished configuring their hotkey.
> 
> I think this is an epic solution! But we should also beware that if the user closes the window (X or altf4) while the edit dialog is opened, it should also recover the switch hotkey ability. Because if someone clicks on the edit button, closes the window with the X, he kinda loses his hotkey abilities.

There's two ways to do this. We should discuss on which ones we want to use

## 1. Frontend only
Ensure that if we are in that state and the frontend is terminated with alt+f4 we re-enable the setting

Advantages:
- less changes
- easier to implement

Disadvantages:
- can potentially fail
- requires weird editing of the config file

## 2. Backend support for temporarily disabling hotkeys
We add a command, for example `--enter-hotkey-config-mode` and `--exit-hotkey-config-mode`
- The enter-config mode stops the hotkey listener for, let's say, 60 seconds and has to be refreshed by the frontend every time (check-in)
- The exit-config mode re-enables the hotkey functionality, but only if the feature is turned on in the config

We can then resend the check-in command every 50 seconds once the dialog is opened. If anything happens to the app, hotkeys get automatically re-enabled (if they are enabled in the config)
If the dialog is closed, we call the exit-mode function explicity, also re-enabling hotkeys (if they are enabled in the config).

With disabling hotkeys, I probably think it's a good idea to add a variable to the GlobalState that gets set if hotkeys should be temporarily disabled, and then just return in the wndproc function that receives the hotkey events.

Advantages:
- config file is not touched, can be done in-place
- never fails, automatically restores hotkey state if check-ins from the frontend stop

Disadvantages:
- more complex to implement
- requires both frontend and backend changes

Spiritreader:
> I can add another (strange) use case, for the _current_ master branch:
> 
>     * When I press (e.g.) Win + NumPad5; ADM hotkey input says "Win NumPad5" ✔️
> 
>     * When I press Shift + {a NumPad number}, the key-function that gets returned is what it would be if **NumLock** would be off. So
>       
>       * Pressing Win + Shift + NumPad1 shows "Win End" ✖️
>       * Pressing Win + Shift + NumPad2 shows "Win + Down" which doesn't get picked up by the ADM input window at all. It minimizes the current window.
> 
> 
> Using a **MS wireless keyboard 3050**, in case it matters.

@Jay-o-Way oh okay, so the current state is broken in stable? How does the described behavior behave in this PR?

Jay-o-Way:
> @Jay-o-Way oh okay, so the current state is broken in stable? How does the described behavior behave in this PR?

Mostly the same (updated earlier comment too)
Also noticed that pressing Win+Shift+1 acts like Win+1, which opens the first pinned app on the taskbar.
Hotkeys are extremely difficult to get right.

Spiritreader:
> Hotkeys are extremely difficult to get right.

Hm yes. I think there will always be conflicts with Windows default behavior. The question is if we want to add some sort of sanity-check in the frontend which prevents hotkeys from being set that are a potential conflict with Windows-internal hotkeys that cannot be changed.

ChenYiLins:
> > @Jay-o-Way oh okay, so the current state is broken in stable? How does the described behavior behave in this PR?
> 
> Mostly the same (updated earlier comment too) Also noticed that pressing Win+Shift+1 acts like Win+1, which opens the first pinned app on the taskbar. Hotkeys are extremely difficult to get right.

Yes, I also found this problem, which existed in the previous 10 versions of ADM. But I think we can actually try to improve the way to get hot keys to solve this problem. (This is absolutely possible)

ChenYiLins:
> @ChenYiLins If this is an issue, we could temporarily disable the hotkey functionality by setting the config option in the yaml file to false and sving it when opening the edit dialog. Then afterwards, we could re-enable it when the user has finished configuring their hotkey.

Great! (Why didn't we think of before hahah)

ChenYiLins:
> ## 2. Backend support for temporarily disabling hotkeys

Let me try to see if there is a simpler solution. (ง •_•)ง

ChenYiLins:
## Code quality: Change hotkey acquisition to system hook

### Description

I learned [something](https://github.com/microsoft/PowerToys/blob/main/src/modules/cmdpal/Microsoft.CmdPal.UI/Controls/ShortcutControl/HotkeySettingsControlHook.cs) from Powertoys.

I have to say that this old method is really old! But it is definitely effective enough! ! !
Now that I've started using [`CallNextHookEx`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-callnexthookex), many of our worries have been solved at once. System hotkeys will no longer be unavailable, and we don't need to turn off the hotkey function, because hook won't touch this series of problems!

Armin2208:
> ## Code quality: Change hotkey acquisition to system hook
> ### Description
> 
> I learned [something](https://github.com/microsoft/PowerToys/blob/main/src/modules/cmdpal/Microsoft.CmdPal.UI/Controls/ShortcutControl/HotkeySettingsControlHook.cs) from Powertoys.
> 
> I have to say that this old method is really old! But it is definitely effective enough! ! ! Now that I've started using [`CallNextHookEx`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-callnexthookex), many of our worries have been solved at once. System hotkeys will no longer be unavailable, and we don't need to turn off the hotkey function, because hook won't touch this series of problems!

That's amazing!

ChenYiLins:
@Spiritreader Can pulling branches like this effectively solve the conflict problem?

Spiritreader:
@ChenYiLins i need to check after wer merge into the v11.1 branch how it behaves when i attempt to sync with master

Spiritreader:
> ## Code quality: Change hotkey acquisition to system hook
> ### Description
> I learned [something](https://github.com/microsoft/PowerToys/blob/main/src/modules/cmdpal/Microsoft.CmdPal.UI/Controls/ShortcutControl/HotkeySettingsControlHook.cs) from Powertoys.
> 
> I have to say that this old method is really old! But it is definitely effective enough! ! ! Now that I've started using [`CallNextHookEx`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-callnexthookex), many of our worries have been solved at once. System hotkeys will no longer be unavailable, and we don't need to turn off the hotkey function, because hook won't touch this series of problems!

How does that prevent the Service from executing a hotkey command when a hotkey is being rebound to another function?

ChenYiLins:
> How does that prevent the Service from executing a hotkey command when a hotkey is being rebound to another function?

No, in fact, the `RegisterHotKey` we use is not always successful, and it is not an exception, so it will not be captured. For example, when the hotkey uses `Win + Shift + Number1`:

In the **HotkeyHandler.cs**
```c#
            Service.Invoke(new Action(() =>
            {
                var result=RegisterHotKey(Service.Handle, 0, modifiers, keyCode);
                if(result)
                {
                    System.Diagnostics.Debug.WriteLine("Register");
                }
                else
                {
                    System.Diagnostics.Debug.WriteLine("Failed to register");
                }
            }));
```

We will get:

```
Failed to register
```

In this regard, I think a better solution is to remind the user that this hotkey has been occupied by the system at the front end, or to forcibly use the **Low-Level Keyboard Hook** at the back end, which can prevent some system hotkeys.

Spiritreader:
@ChenYiLins What you say is correct of course. Some sys hotkeys can't bet set, and I also don't think we should attempt to override, so a message would be a good idea, yes.

But what I'm referring to in my previous comment is if a user sets `alt+shift+i` for, let's say "toggle light" and then also attempts to set `alt+shift+i` for a different function. Then the backend will execute the hotkey while a user attempt so remap.

ChenYiLins:
> @ChenYiLins What you say is correct of course. Some sys hotkeys can't bet set, and I also don't think we should attempt to override, so a message would be a good idea, yes.
> 
> But what I'm referring to in my previous comment is if a user sets `alt+shift+i` for, let's say "toggle light" and then also attempts to set `alt+shift+i` for a different function. Then the backend will execute the hotkey while a user attempt so remap.

Ah, I misunderstood. Sorry. But in fact, this will not happen, because the priority of hook is higher than the hot key function of the back end.

PixPin_2025-10-31_23-21-24

Spiritreader:
> Ah, I misunderstood. Sorry. But in fact, this will not happen, because the priority of hook is higher than the hot key function of the back end.

Oh okay, awesome!
From my end, this looks good and we can start testing in beta. If issues come up we can improve as we go

Spiritreader (approved):

Issue #1102, opened 2025-10-14
[Bug] JumpList doesn't work


Armin2208:
### Description

Jump Lists are still part of AutoDarkMode App, but they don't work anymore. They just open the app. They don't run the command.

### Expected Behavior

ADM shouldn't open while using jump list, the app should just send a command to service, just like Auto Dark Mode X

### Log Data

-

### Commit Hash, Version and Windows Build

- Commit: `5678573e`
- Service/App: `11.0.0.54`
- Updater: `4.0.3`
- Shell: `1.3.3.0`
- .Net: `9.0.8`
- Windows: `26100.6584`
- Arch: `X64`

Jay-o-Way:
Hm, do we know where it broke?

Spiritreader:
> Hm, do we know where it broke?

Yes we do. Jump list functionality was removed in Winui3. The entire code is gone.
I tried to re-implement it but the winui3 lifecycles make it impossible to use the UWP bindings, so we need to completely rewrite this it seems like.

The problem is that CLI args passed to Winui3 via the jumplist no longer get picked up.

PR #1135, opened 2025-10-27
Fix: JumpList doesn't work

ChenYiLins:
### Decription

With the analysis of parameters at startup, JumpList works as usual.

Close: #1102

### DIFF
--- a/AutoDarkModeApp/App.xaml.cs
+++ b/AutoDarkModeApp/App.xaml.cs
@@ -5,6 +5,7 @@ using AutoDarkModeApp.Models;
 using AutoDarkModeApp.Services;
 using AutoDarkModeApp.ViewModels;
 using AutoDarkModeApp.Views;
+using AutoDarkModeComms;
 using AutoDarkModeLib;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -118,10 +119,28 @@ public partial class App : Application
     {
         base.OnLaunched(args);
 
+        // Handle JumpListCommand
+        var arguments = Environment.GetCommandLineArgs();
+        if (arguments.Length > 1)
+        {
+            new PipeClient().SendMessageAndGetReply(arguments[1]);
+            Environment.Exit(-1);
+        }
+
+        // Set App and Svc language
         Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = await LanguageHelper.GetDefaultLanguageAsync();
-        var admConfigBuilder = AdmConfigBuilder.Instance();
-        admConfigBuilder.Load();
-        admConfigBuilder.Config.Tunable.UICulture = LanguageHelper.SelectedLanguageCode; // save before activation SVC (think of first-launch scenario)
+        var builder = AdmConfigBuilder.Instance();
+        builder.Load();
+        try
+        {
+            builder.Config.Tunable.UICulture = LanguageHelper.SelectedLanguageCode; // save before activation SVC (think of first-launch scenario)
+            builder.Save();
+        }
+        catch (Exception)
+        {
+            // We can't show a dialog here 
+            //await App.GetService<IErrorService>().ShowErrorMessage(ex, App.MainWindow.Content.XamlRoot, "Startup", "Failed to force-set Svc language");
+        }
 
         MainWindow = new MainWindow();
 
--- a/AutoDarkModeApp/Services/ActivationService.cs
+++ b/AutoDarkModeApp/Services/ActivationService.cs
@@ -213,8 +213,6 @@ public class ActivationService(ILocalSettingsService localSettingsService, INavi
             var jumpList = await JumpList.LoadCurrentAsync();
 
             jumpList.Items.Clear();
-            Debug.WriteLine("Adding jumplist items");
-            Debug.WriteLine($"Current override: {Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride}");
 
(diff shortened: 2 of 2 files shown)


Spiritreader:
Oh, I did try this exact same method to read the CLI: `Environment.GetCommandLineArgs();`

But it didn't work. I must've missed something or the build didn't go through as expected.

Spiritreader (approved):

ChenYiLins:
> Oh, I did try this exact same method to read the CLI: `Environment.GetCommandLineArgs();`
> 
> But it didn't work. I must've missed something or the build didn't go through as expected.

Hhhh. Maybe: _The first element in the array contains the file name of the executing program. If the file name is not available, the first element is equal to [String.Empty](https://learn.microsoft.com/en-us/dotnet/api/system.string.empty?view=net-9.0#system-string-empty)._

Doc: https://learn.microsoft.com/en-us/dotnet/api/system.environment.getcommandlineargs?view=net-9.0#system-environment-getcommandlineargs

Spiritreader:
> > Oh, I did try this exact same method to read the CLI: `Environment.GetCommandLineArgs();`
> > But it didn't work. I must've missed something or the build didn't go through as expected.
> 
> Hhhh. Maybe: _The first element in the array contains the file name of the executing program. If the file name is not available, the first element is equal to [String.Empty](https://learn.microsoft.com/en-us/dotnet/api/system.string.empty?view=net-9.0#system-string-empty)._
> 
> Doc: https://learn.microsoft.com/en-us/dotnet/api/system.environment.getcommandlineargs?view=net-9.0#system-environment-getcommandlineargs

The problem I had was that the list of cli args was always empty, regardless of how i called the exectuable.
But I'm not complaining that it works now :D

Update: Hmm, I tried on my PC on the production build, and it still doesn't work correctly. The app starts, but no jumplist command is sent to the backend.

For Armin jump lists seem to work except for the reset command

Armin2208:
JumpList is working here on my machine! It switches perfectly between dark and light theme. Only the "reset" jump list entry does not work.

But I see the following behavior:

1. Current theme is dark mode, automatic theme switch is enabled.
2. I click on "light theme" on the jump list.
3. Theme switches to light and automatic theme switch is paused until next day.
4. I click again on "dark theme". The theme switch is not paused anymore!

Originally the buttons forced the theme, currently it seems like it only  toggles it kinda.

My opinion: instead of fixing the "reset" button, we should remove it. Maybe the behaviour is then pretty good and user friendly.

grafik

Spiritreader:
Update @ChenYiLins: Jumplists aren't working on my system because I use a modified Windows taskbar. We can ignore this use case because using non standard Windows components that don't follow the API specification correctly is outisde of our scope.

If I use the "normal" Windows 11 taskbar, I can confirm however that that Reset jump list item does indeed not work.

ChenYiLins:
> Update @ChenYiLins: Jumplists aren't working on my system because I use a modified Windows taskbar. We can ignore this use case because using non standard Windows components that don't follow the API specification correctly is outisde of our scope.
> 
> If I use the "normal" Windows 11 taskbar, I can confirm however that that Reset jump list item does indeed not work.

I went to check the code, but the Reset function is actually effective. But it is used to clear the forced application theme.

```c#
            var darkJumpTask = JumpListItem.CreateWithArguments(Command.Dark, "DarkTheme".GetLocalized());
            darkJumpTask.GroupName = "SwitchTheme".GetLocalized();

            var lightJumpTask = JumpListItem.CreateWithArguments(Command.Light, "LightTheme".GetLocalized());
            lightJumpTask.GroupName = "SwitchTheme".GetLocalized();

            var resetJumpTask = JumpListItem.CreateWithArguments(Command.NoForce, "Reset".GetLocalized());
            resetJumpTask.GroupName = "SwitchTheme".GetLocalized();
```

I don't know if this means that the original code is wrong, but we should actually set the forced application theme.╰(*°▽°*)╯

Spiritreader:
> > Update @ChenYiLins: Jumplists aren't working on my system because I use a modified Windows taskbar. We can ignore this use case because using non standard Windows components that don't follow the API specification correctly is outisde of our scope.
> > If I use the "normal" Windows 11 taskbar, I can confirm however that that Reset jump list item does indeed not work.
> 
> I went to check the code, but the Reset function is actually effective. But it is used to clear the forced application theme.
> 
> ```cs
>             var darkJumpTask = JumpListItem.CreateWithArguments(Command.Dark, "DarkTheme".GetLocalized());
>             darkJumpTask.GroupName = "SwitchTheme".GetLocalized();
> 
>             var lightJumpTask = JumpListItem.CreateWithArguments(Command.Light, "LightTheme".GetLocalized());
>             lightJumpTask.GroupName = "SwitchTheme".GetLocalized();
> 
>             var resetJumpTask = JumpListItem.CreateWithArguments(Command.NoForce, "Reset".GetLocalized());
>             resetJumpTask.GroupName = "SwitchTheme".GetLocalized();
> ```
> 
> I don't know if this means that the original code is wrong, but we should actually set the forced application theme.╰(_°▽°_)╯

I know what's happening:
The original intent of the jumplist was to set the "ForceLight" and "ForceDark" flags, but since we're now sending Command.Dark and Command.Light that no longer applies. The commands for the jumplist need to be reworked anyway to reflect the new style of ADM's postpone system.

The force flags are legacy behavior in the first place.

Everything is actually working as intended, we can remove the "Reset" jumplist so we only have the Light and Dark options available.

What's happening currently is that if by system-time we should be in Dark mode, and you select Light as jump-list option,ADM switches to light and postpones, and vice versa, which is how it should be. Changing it back while it's postponed will correctly clear the postpone, so that also works as expected.

PR #1136, opened 2025-10-27
Code quality and fix: About ColorizationPage

ChenYiLins:
### Description

- Fix the Bug of ColorizationPage crash
- Hide accent color border(like a button) when automatic is selected


Spiritreader (approved):
good catch, thanks!

PR #1151, opened 2025-11-11
Tweak donation page icons

Jay-o-Way:
Noticed that the PayPal icon was a bit "blocky". Removed the font file.

Before:
before

After:
after

Note: I created new, scaled files to keep the old ones for reference. Let me know if it is better to overwrite the originals?


Armin2208 (approved):
thats good

PR #1153, opened 2025-11-12
Code quality: Code cleanup

ChenYiLins:
### Description

In this PR, there won't be any code logic changes and function updates, just simply run code cleanup on the back end.

This will lay a good foundation for future back end changes (code review will not become more difficult). If the time is right, it is also a good idea to use sharpier (the code style will be more consistent).

### Supplement

And #1041.

### DIFF
--- a/AutoDarkModeSvc/Core/ComponentManager.cs
+++ b/AutoDarkModeSvc/Core/ComponentManager.cs
@@ -180,9 +180,9 @@ class ComponentManager
     }
 
     private static void AddComponentAndGetDwmInfo(
-        ISwitchComponent c, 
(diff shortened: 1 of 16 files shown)


Spiritreader:
Hi!

Those changes were fine to review as it's only small format changes!

I'm planning to apply your sharpier configuration to the entire repo at some point.

After that we can start making fine grained changes, but I think all of the changes that modify the current structure need a proposal first. 

There's a lot to improve at the backend, as it carries all of the tech debt of me learning how to become a C# programmer 😆
But windows theme switching is also really delicate!

Spiritreader (approved):

PR #1157, opened 2025-11-23
Fix and Code quality: About WallpaperPickerPage

ChenYiLins:
### Description

- Fix bug caused by null values
- Unified design with ColorizationPage

### Sreenshots

image

image

### DIFF
--- a/AutoDarkModeApp/Views/WallpaperPickerPage.xaml
+++ b/AutoDarkModeApp/Views/WallpaperPickerPage.xaml
@@ -46,7 +46,7 @@
                 </controls:SettingsCard>
 
                 <controls:SettingsExpander
-                    Header="{helpers:ResourceString Name=PickMode}"
+                    Header="{helpers:ResourceString Name=Mode}"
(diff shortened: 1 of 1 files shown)


Armin2208 (approved):
finally

AutoDarkModeApp/Strings/en-us/Resources.resw
```diff
@@ -1027,4 +1027,7 @@ On rare occasions, Windows may emit a beeping sound when the mouse is moved duri
   <data name="AccentApplyTaskbar" xml:space="preserve">
     <value>Accent color for the taskbar</value>
   </data>
+  <data name="PickMode" xml:space="preserve">
+    <value>Select a mode</value>
```
Jay-o-Way: I say we could or should simply use the string "Mode". Short and simple 🙂

ChenYiLins: Thank you for your advice! It has been changed! ( •̀ ω •́ )✧

Spiritreader:
Thank you @ChenYiLins!

PR #1164, opened 2025-12-24
csproj: add info to catch git issues

Jay-o-Way:
add info to csproj to catch issues when git may not be available.


ChenYiLins:
@Spiritreader @Armin2208 Is this OK? φ(*￣0￣)

Armin2208:
i guess

Spiritreader:
Not quite sure why someone would clone a git repo and then not have git, but sure, go ahead.

Issue #1070, opened 2025-09-18
[Feature] Switch light/dark mode according to the ambient light around


tomzi8571:
I couldn't find this feature yet in your app, so excuse me, if this is redundent, but if it isn't I guess you probably already thought about it, anyways t here is my feature request anyways 😉:

I often change places where I am working or developing on my Laptop. Sometimes when I work remote, in my rooom, or if it is really nice outside, then on the porch.

However when I am outside, working in dark mode is really difficult to read and work, so I switch to light mode and full brightness, when I am inside then it is more comfortable for my eyes to work in dark mode.

So I keep switching manually between dark and light mode.

You can guess where this is going, right? I have researched, that on windows 11 we already have in the Settings -> System -> Display -> Brightness, the option to 'change brightness when the lighting changes' and 'Change brightness based on the content' but it **doesn't** provide **'Change Dark/Light mode when lighting changes'**. I guess they might add this feature add some point - But you don't really now when and if they will keep it.

So, I think that would be a really helpful addtion and great feature for this awesome app you created. 

** Implementation hints'
Since I am home in the Enterprise Java world and don't know a lot about UWP Apps I was interested and had really fun to ask Github Copilot regarding your project and how it would implement it... that's the answer below 😂 It's probably not very helpful, though. Anyways regarding the implementation strategy, I would probably adapt, that there is no 'hard' switch at a certain level of 'lux', but a cross over feature,  like Github Copilot interestingly already suggested, as you can see below.

What do you think?

**Github CoPilot**

**Where to Change the Current Code**
Key Areas:
- Core Logic: Add a new scheduler/service for ambient light detection.
- UI: Add options in settings UI.
- Theme Switch Logic: Integrate ambient-based switching as one of the triggers.

**Example Pseudocode (C#/.NET)**

 ``` C#
using Windows.Devices.Sensors;

public class AmbientLightMonitor
{
    private LightSensor _sensor;
    private double _lastLux = 0;
    private const double DarkThreshold = 40;
    private const double LightThreshold = 80;
    private bool _isDarkMode = false;

    public AmbientLightMonitor()
    {
        _sensor = LightSensor.GetDefault();
        if (_sensor != null)
        {
            _sensor.ReadingChanged += OnReadingChanged;
        }
    }

    private void OnReadingChanged(LightSensor sender, LightSensorReadingChangedEventArgs args)
    {
        var lux = args.Reading.IlluminanceInLux;
[...]

Issue #15, opened 2019-01-18
backend support: more ways to switch theme


Armin2208:
Switching theme...
- when lowering screen brightness.
- when ambient sensor detects a dark environment.
- when energy saver is on.
- when pressing a keyboard shortcut.

A background task is needed for this and could be introduced in version 3.

Elia0107:
It would be nice if the theme can change also when the wallpaper is predominantly dark or light. Like CullaX on KDE for Linux. Of course this need background task, but the windows wallpaper can be easily found in "%AppData%\Microsoft\Windows\Themes\CachedFiles".

Armin2208:
Sounds nice, but I will not implement this. Sorry. I don't think that's necessary.

Spiritreader:
I think the background service project can be combined with #26.

When the background service is started, it would be useful to show a tray/notification icon that indicates that it's running.

Then it would also be possible to configure right click entries for switching themes there.

I'm very interested in contributing on this milestone together with @utkucanturkan.
Did you already have a fixed plan on implementing a background service?

If not, we have briefly talked about this and came to the conclusion that a highly modularized approach would be very beneficial to the ability to maintain and create new background service modules.

### Idea
A background service is created that also spawns a configurable tray icon.
Since we already have a tray icon this way, we can also provide options like launching the interface from there 

The service offers an interface that provides the ability to register and deregister modules. It would also take care of creating and destroying instances whenever necessary.
Registered modules run on an interruptable timer which should prove to be very resource efficient.

Each module implements a standardized interface that exposes methods for periodically polling a module.

Armin2208:
@Spiritreader can we chat about that, in telegram?

skaiteo:
Any plans to implement the brightness trigger?

Spiritreader:
> Any plans to implement the brightness trigger? Say, when the brightness goes below a certain level.

It's still planned but probably not anytime soon as that's a feature that needs to be implemented from scratch. 
There are a lot of other issues and improvements that take precedence right now.
Win11 has been a major pain to integrate so far, and there are still outstanding issues that have higher priority.

skaiteo:
Roger that, thanks for the update!

ssssamstrup:
I think this will be hard to implement as well, unless automatic dimming is disabled by the user/program. In certain environments it'll be close to impossible to work in a way that doesn't trigger the program to constantly switch. 
[...]

PR #1168, opened 2026-01-12
feat(governor): add ambient light sensor theme switching

BYK:
## Summary

Implements #1070 - Switch light/dark mode according to ambient light levels detected by the device's light sensor.

image

### New Governor
- Added `AmbientLightGovernor` using `Windows.Devices.Sensors.LightSensor` API
- Hysteresis logic with configurable thresholds to prevent rapid switching
- Theme updates immediately when sensor readings cross thresholds

### Configuration
- Added `AmbientLight` config class with `DarkThreshold` (default: 40 lux) and `LightThreshold` (default: 80 lux)
- Thresholds enforce minimum 5 lux gap between dark and light values

### UI Enhancements
- New "Ambient light sensor" option in switch trigger modes
- Real-time lux reading display with descriptive labels (e.g. "45 lux — Dimly lit room")
- Visual sliders + number inputs for threshold configuration
- Expandable reference guide showing typical lux values for different environments
- Immediate theme preview when adjusting thresholds

### Refactoring
- Renamed `TimeSourceMode` to `SwitchTriggerMode` for semantic clarity (ambient light isn't time-based)

## Notes

⚠️ **This PR was vibe coded with AI assistance.** I don't have much experience with .NET codebases, but local testing worked quite well:
- Theme changes immediately when thresholds are adjusted (no need to wait for light level changes)
- The 5 lux minimum threshold gap prevents invalid configurations
- Real-time lux display helps users pick appropriate thresholds for their environment

Happy to address any feedback or concerns!

### DIFF
--- a/AutoDarkModeApp/Views/TimePage.xaml
+++ b/AutoDarkModeApp/Views/TimePage.xaml
@@ -15,25 +15,6 @@
         <helpers:EnumToIndexConverter x:Key="EnumToIndexConverter" />
         <helpers:EnumToVisibilityConverter x:Key="EnumToVisibilityConverter" />
         <helpers:EnumToBooleanConverter x:Key="EnumToBooleanConverter" />
-
-        <!--  Styles for ambient light slider tick marks  -->
-        <Style x:Key="TickMarkStackPanelStyle" TargetType="StackPanel">
-            <Setter Property="HorizontalAlignment" Value="Right" />
-            <Setter Property="Spacing" Value="2" />
-            <Setter Property="Margin" Value="0,0,-12,0" />
-            <Setter Property="Width" Value="24" />
-        </Style>
-        <Style x:Key="TickMarkBorderStyle" TargetType="Border">
-            <Setter Property="Width" Value="1" />
-            <Setter Property="Height" Value="6" />
-            <Setter Property="Background" Value="{ThemeResource TextFillColorTertiaryBrush}" />
-            <Setter Property="HorizontalAlignment" Value="Center" />
-        </Style>
-        <Style x:Key="TickMarkTextBlockStyle" TargetType="TextBlock">
-            <Setter Property="FontSize" Value="10" />
-            <Setter Property="Foreground" Value="{ThemeResource TextFillColorTertiaryBrush}" />
-            <Setter Property="HorizontalAlignment" Value="Center" />
-        </Style>
     </Page.Resources>
 
     <ScrollViewer x:Name="PageScrollViewer">
@@ -337,22 +318,78 @@
                                                             <RowDefinition Height="Auto" />
                                                             <RowDefinition Height="Auto" />
                                                         </Grid.RowDefinitions>
-                                                        <TextBlock Grid.Row="0" Grid.Column="0" Text="&lt; 1 lux" />
-                                                        <TextBlock Grid.Row="0" Grid.Column="1" Text="{helpers:ResourceString Name=AmbientLightRef_Moonlight}" Foreground="{ThemeResource TextFillColorSecondaryBrush}" />
-                                                        <TextBlock Grid.Row="1" Grid.Column="0" Text="1 - 10 lux" />
-                                                        <TextBlock Grid.Row="1" Grid.Column="1" Text="{helpers:ResourceString Name=AmbientLightRef_VeryDark}" Foreground="{ThemeResource TextFillColorSecondaryBrush}" />
-                                                        <TextBlock Grid.Row="2" Grid.Column="0" Text="10 - 50 lux" />
-                                                        <TextBlock Grid.Row="2" Grid.Column="1" Text="{helpers:ResourceString Name=AmbientLightRef_DimRoom}" Foreground="{ThemeResource TextFillColorSecondaryBrush}" />
-                                                        <TextBlock Grid.Row="3" Grid.Column="0" Text="50 - 150 lux" />
-                                                        <TextBlock Grid.Row="3" Grid.Column="1" Text="{helpers:ResourceString Name=AmbientLightRef_LivingRoom}" Foreground="{ThemeResource TextFillColorSecondaryBrush}" />
-                                                        <TextBlock Grid.Row="4" Grid.Column="0" Text="150 - 400 lux" />
-                                                        <TextBlock Grid.Row="4" Grid.Column="1" Text="{helpers:ResourceString Name=AmbientLightRef_Office}" Foreground="{ThemeResource TextFillColorSecondaryBrush}" />
-                                                        <TextBlock Grid.Row="5" Grid.Column="0" Text="400 - 1000 lux" />
-                                                        <TextBlock Grid.Row="5" Grid.Column="1" Text="{helpers:ResourceString Name=AmbientLightRef_Overcast}" Foreground="{ThemeResource TextFillColorSecondaryBrush}" />
-                                                        <TextBlock Grid.Row="6" Grid.Column="0" Text="1000 - 10000 lux" />
-                                                        <TextBlock Grid.Row="6" Grid.Column="1" Text="{helpers:ResourceString Name=AmbientLightRef_DaylightShade}" Foreground="{ThemeResource TextFillColorSecondaryBrush}" />
-                                                        <TextBlock Grid.Row="7" Grid.Column="0" Text="&gt; 10000 lux" />
-                                                        <TextBlock Grid.Row="7" Grid.Column="1" Text="{helpers:ResourceString Name=AmbientLightRef_DirectSun}" Foreground="{ThemeResource TextFillColorSecondaryBrush}" />
+                                                        <TextBlock
+                                                            Grid.Row="0"
+                                                            Grid.Column="0"
+                                                            Text="&lt; 1 lux" />
+                                                        <TextBlock
+                                                            Grid.Row="0"
+                                                            Grid.Column="1"
+                                                            Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                                            Text="{helpers:ResourceString Name=AmbientLightRef_Moonlight}" />
+                                                        <TextBlock
+                                                            Grid.Row="1"
+                                                            Grid.Column="0"
+                                                            Text="1 - 10 lux" />
+                                                        <TextBlock
+                                                            Grid.Row="1"
+                                                            Grid.Column="1"
+                                                            Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                                            Text="{helpers:ResourceString Name=AmbientLightRef_VeryDark}" />
+                                                        <TextBlock
+                                                            Grid.Row="2"
+                                                            Grid.Column="0"
+                                                            Text="10 - 50 lux" />
+                                                        <TextBlock
+                                                            Grid.Row="2"
+                                                            Grid.Column="1"
+                                                            Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                                            Text="{helpers:ResourceString Name=AmbientLightRef_DimRoom}" />
+                                                        <TextBlock
+                                                            Grid.Row="3"
+                                                            Grid.Column="0"
+                                                            Text="50 - 150 lux" />
+                                                        <TextBlock
+                                                            Grid.Row="3"
+                                                            Grid.Column="1"
+                                                            Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                                            Text="{helpers:ResourceString Name=AmbientLightRef_LivingRoom}" />
+                                                        <TextBlock
+                                                            Grid.Row="4"
+                                                            Grid.Column="0"
+                                                            Text="150 - 400 lux" />
+                                                        <TextBlock
+                                                            Grid.Row="4"
+                                                            Grid.Column="1"
+                                                            Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                                            Text="{helpers:ResourceString Name=AmbientLightRef_Office}" />
+                                                        <TextBlock
+                                                            Grid.Row="5"
+                                                            Grid.Column="0"
+                                                            Text="400 - 1000 lux" />
+                                                        <TextBlock
+                                                            Grid.Row="5"
+                                                            Grid.Column="1"
+                                                            Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                                            Text="{helpers:ResourceString Name=AmbientLightRef_Overcast}" />
+                                                        <TextBlock
+                                                            Grid.Row="6"
+                                                            Grid.Column="0"
+                                                            Text="1000 - 10000 lux" />
+                                                        <TextBlock
+                                                            Grid.Row="6"
+                                                            Grid.Column="1"
+                                                            Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                                            Text="{helpers:ResourceString Name=AmbientLightRef_DaylightShade}" />
+                                                        <TextBlock
+                                                            Grid.Row="7"
+                                                            Grid.Column="0"
+                                                            Text="&gt; 10000 lux" />
+                                                        <TextBlock
+                                                            Grid.Row="7"
+                                                            Grid.Column="1"
+                                                            Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                                            Text="{helpers:ResourceString Name=AmbientLightRef_DirectSun}" />
                                                     </Grid>
                                                 </Flyout>
                                             </Button.Flyout>
@@ -430,9 +467,9 @@
                                             Minimum="0"
                                             RangeEnd="{x:Bind ViewModel.RangeEnd, Mode=TwoWay}"
                                             RangeStart="{x:Bind ViewModel.RangeStart, Mode=TwoWay}"
-                                            ThumbDragStarted="AmbientLightRangeSelector_ThumbDragStarted"
+                                            StepFrequency="1"
                                             ThumbDragCompleted="AmbientLightRangeSelector_ThumbDragCompleted"
-                                            StepFrequency="1" />
+                                            ThumbDragStarted="AmbientLightRangeSelector_ThumbDragStarted" />
                                     </Grid>
 
                                     <!--  Indicator showing current lux level on the slider  -->
--- a/AutoDarkModeApp/Styles/CustomStyles.xaml
+++ b/AutoDarkModeApp/Styles/CustomStyles.xaml
@@ -20,6 +20,25 @@
         <Setter Property="Spacing" Value="{StaticResource SettingsCardSpacing}" />
     </Style>
 
+    <!--  Styles for ambient light slider tick marks  -->
+    <Style x:Key="TickMarkStackPanelStyle" TargetType="StackPanel">
+        <Setter Property="HorizontalAlignment" Value="Right" />
+        <Setter Property="Spacing" Value="2" />
+        <Setter Property="Margin" Value="0,0,-12,0" />
+        <Setter Property="Width" Value="24" />
+    </Style>
+    <Style x:Key="TickMarkBorderStyle" TargetType="Border">
+        <Setter Property="Width" Value="1" />
+        <Setter Property="Height" Value="6" />
+        <Setter Property="Background" Value="{ThemeResource TextFillColorTertiaryBrush}" />
+        <Setter Property="HorizontalAlignment" Value="Center" />
+    </Style>
+    <Style x:Key="TickMarkTextBlockStyle" TargetType="TextBlock">
+        <Setter Property="FontSize" Value="10" />
+        <Setter Property="Foreground" Value="{ThemeResource TextFillColorTertiaryBrush}" />
+        <Setter Property="HorizontalAlignment" Value="Center" />
+    </Style>
+
     <!--  Style of a section header  -->
     <Style
         x:Key="SettingsSectionHeaderTextBlockStyle"
@@ -42,22 +61,22 @@
             <!--  BreadcrumbBar style like Win.11 Settings  -->
             <StaticResource x:Key="BreadcrumbBarForegroundBrush" ResourceKey="TextFillColorSecondaryBrush" />
             <StaticResource x:Key="BreadcrumbBarHoverForegroundBrush" ResourceKey="TextFillColorPrimaryBrush" />
-            <!-- Theme sensitive icons -->
+            <!--  Theme sensitive icons  -->
             <BitmapImage x:Key="PayPalLogo" UriSource="/Assets/PayPal_Logo64.png" />
             <BitmapImage x:Key="GithubLogo" UriSource="/Assets/Github_Dark_Logo64.png" />
-            <BitmapImage x:Key="TelegramLogo" UriSource="/Assets/Telegram_Logo24.png"/>
+            <BitmapImage x:Key="TelegramLogo" UriSource="/Assets/Telegram_Logo24.png" />
         </ResourceDictionary>
-        
+
         <ResourceDictionary x:Key="Dark">
             <!--  BreadcrumbBar style like Win.11 Settings  -->
             <StaticResource x:Key="BreadcrumbBarForegroundBrush" ResourceKey="TextFillColorSecondaryBrush" />
             <StaticResource x:Key="BreadcrumbBarHoverForegroundBrush" ResourceKey="TextFillColorPrimaryBrush" />
-            <!-- Theme sensitive icons -->
+            <!--  Theme sensitive icons  -->
             <BitmapImage x:Key="PayPalLogo" UriSource="/Assets/PayPal_Logo64.png" />
             <BitmapImage x:Key="GithubLogo" UriSource="/Assets/Github_White_Logo64.png" />
-            <BitmapImage x:Key="TelegramLogo" UriSource="/Assets/Telegram_Logo24.png"/>
+            <BitmapImage x:Key="TelegramLogo" UriSource="/Assets/Telegram_Logo24.png" />
         </ResourceDictionary>
-        
+
         <ResourceDictionary x:Key="HighContrast">
             <!--  BreadcrumbBar style like Win.11 Settings  -->
             <StaticResource x:Key="BreadcrumbBarForegroundBrush" ResourceKey="TextFillColorSecondaryBrush" />


Jay-o-Way:
Ooh, this looks cool! Apparently, I don't have a light sensor, but that doesn't keep me from testing things. So far, I can tell I would definitely do things different in the settings UI. Quick first comments:

- The first SettingsCard with "Current light level" doesn't have any content for me, because my computer doesn't have a sensor. Even if this should only be accessible when the RadioButton is selected, it would be extra nice to have some kind of default/fallback text for info.
- Can we move the (new) settings inside the larger area above? Just like settings for other switch options.
- What's the maximum lux value? **500** (in the Slider) or **1000** (in the NumberBox) ?
- Some more small UI bits...

BYK:
@Jay-o-Way thanks so much for the quick turn around and encouragement! Made some updates based on your feedback and updated the screenshot in the PR description.

LMK what you think now

Jay-o-Way:
Thanks @BYK, that was fast! Nice improvements so far! 👍🏻

A few more thoughts for now:

- Both the sliders may have a maximum width, for example **500**. (done)
- Interaction between the two sliders is...interesting. @ChenYiLins I remember you used `args.handled = false` here and there to revert unallowed values. Would that thing be helpful here, in `OnAmbientLightDarkThresholdChanged`? One Slider acts as a limiter for the other Slider. Maybe there's another UX solution? Possibly Community Toolkit's **RangeSelector** control?
- The "legend" or guide is nice to have! 😄 But this doesn't really feel like it's in-line with Windows UX. A subtle **Button** with a Flyout, or a [TeachingTip](https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.teachingtip?view=windows-app-sdk-1.8) would be a nice solution.

AutoDarkModeApp/Views/TimePage.xaml
```diff
@@ -187,7 +188,171 @@
...
+                                        Style="{StaticResource BodyTextBlockStyle}"
+                                        Text="{x:Bind ViewModel.CurrentLuxDescription, Mode=OneWay}" />
+                                </Grid>
+
+                                <Border Style="{StaticResource DividerBorderStyle}" />
+
+                                <!--  Threshold Configuration  -->
+                                <TextBlock
+                                    Style="{StaticResource CaptionTextBlockStyle}"
+                                    Foreground="{ThemeResource TextFillColorSecondaryBrush}"
```
Jay-o-Way:
```suggestion
                                <TextBlock
                                    Style="{StaticResource BodyTextBlockStyle}"
```

Jay-o-Way (commented):
A few ideas

Jay-o-Way:
How do you (all) feel about this?
image

AutoDarkModeApp/Views/TimePage.xaml
```diff
@@ -30,11 +30,12 @@
...
                         <StackPanel Spacing="12">
-                            <RadioButtons SelectedIndex="{x:Bind ViewModel.SelectedTimeSource, Converter={StaticResource EnumToIndexConverter}, Mode=TwoWay}">
+                            <RadioButtons SelectedIndex="{x:Bind ViewModel.SelectedTriggerMode, Converter={StaticResource EnumToIndexConverter}, Mode=TwoWay}">
                                 <RadioButton x:Name="RadioButtonCustomTimes" Content="{helpers:ResourceString Name=CustomHours}" />
                                 <RadioButton x:Name="RadioButtonLocationTimes" Content="{helpers:ResourceString Name=LocationService}" />
                                 <RadioButton x:Name="RadioButtonCoordinateTimes" Content="{helpers:ResourceString Name=LocationGeoService}" />
                                 <RadioButton x:Name="RadioButtonWindowsNightLight" Content="{helpers:ResourceString Name=FollowWindowsNightLight}" />
+                                <RadioButton x:Name="RadioButtonAmbientLight" Content="{helpers:ResourceString Name=AmbientLightSensor}" IsEnabled="{x:Bind ViewModel.AmbientLightSensorAvailable, Mode=OneWay}" />
```
Jay-o-Way: Actually, if something really is **impossible** to select, it can be hidden.
Compare settings for a battery, on a device that does not have a battery. 🔋🚫

Jay-o-Way: Actually, if something really is **impossible** to select, it can be hidden.
Compare settings for a battery, on a device that does not have a battery. 🔋🚫

```suggestion
                                <RadioButton x:Name="RadioButtonAmbientLight" Content="{helpers:ResourceString Name=AmbientLightSensor}" Visibility="{x:Bind ViewModel.AmbientLightSensorAvailable, Mode=OneWay}" />
```

ChenYiLins:
Wow, I! ! This is a very great PR! ! ! Thank you very much for your contribution! ! ♪(^∇^*) Although I noticed that you said that this PR is assisted by AI, I believe that as long as there are enough tests, this will not be a problem!

Jay-o-Way:
@BYK would you prefer to keep making updates yourself, or would you mind if we commit too?
(I could help with a few updates for the UX, including earlier comments)

BYK:
@Jay-o-Way definitely okay you folks taking this over or submitting fixes. Sorry I just didn't have time to come back yet today. Will be available later to see if I need to make more changes.

BYK:
Btw I've been running this build locally since last night and it works really well but cannot speak of battery impact if there's any.

AutoDarkModeApp/Views/TimePage.xaml
Jay-o-Way: FYI: normally, if an option is impossible, it can be hidden. I tried the **Visibility** property, that does hide the control, but it reserves the space it would use. Using `x:Load=` doesn't build. So, until we find a better way, **IsEnabled** is okay.

Jay-o-Way:
@BYK The value of 500 lux is arbitrary (aka a _magic number_). I think 1000 would be a better number, what do you think?

BYK:
@Jay-o-Way sounds good! Here's the LLM's reasoning:

>Looking back, 500 lux was somewhat arbitrary - it was the original slider maximum, and I matched the numberbox to it for consistency.
>But looking at the reference guide:
>Office lighting: 150-400 lux
>Overcast day indoors: 400-1000 lux
>Daylight (shade): 1000-10000 lux
>500 lux might be too limiting for users who work near windows or in brightly lit environments. A threshold of 1000 lux would cover most indoor scenarios including overcast daylight, which is probably more practical.

AutoDarkModeApp/Views/TimePage.xaml
BYK: Would you like me to work on this?

BYK:
@Jay-o-Way made some nice visual updates and switched to a range slider. Update the screenshot in the PR description to show the final stage

Jay-o-Way:
Interesting. Using a logarigthmic scale is smart because it is natural. But it's also complicated. Having these two values that influence eachother makes the UX even more complex.

But there's a problem with the numbers - the ticks don't line up exactly with the NumberBoxes, and the Tooltips use the 0 - 1000 scale.
image

I will try to improve this where I can...

> [!NOTE]
> WORKING ON IT

AutoDarkModeApp/Views/TimePage.xaml
Jay-o-Way: Not at the moment, it's functionally not important. Let's get the UX and code-behind clean, and let the others (Armin and Sam) have a test/review too...

Jay-o-Way:
@tomzi8571 not finished yet, but is this what you had in mind?

BYK:
@Jay-o-Way thanks for checking it out!

- I'm working on an auto-calibrate logic to make this easier for every-day users
- Fixed the ticks not aligning 🤦🏻 
- The tooltips were supposed to be hidden. Tried something else, check again please. They are hidden because of the logarithmic scale: the scale in the range selector is from 0-1000 whereas the translated values are from 0 to 10k and it moves logarithmically as you noticed.

Jay-o-Way:
@BYK I was working on this (in parallel, that's why I said "working on it") and I have a sweeeet solution now. Check this out! (incompatible with your changes though)

RangeSelector in ADM

Are you okay if I push this, when it overwrites what you've done?

BYK:
@Jay-o-Way Oooh I like the new gradient that said I'm not a fan of limiting the scale to 10-1000. For instance I regularly work in environments below 10 lux. Looks like you have kept the logarithmic scale though, right?

Please do not push yet as I am working on more stuff. Hoping for us to find a way to merge these two approaches.

Btw just noticed it says "Time" as the header for the section which obviously needs to change now 😅

Jay-o-Way:
@BYK 

> Oooh I like the new gradient

Thanks!

> I'm not a fan of limiting the scale to 10-1000. For instance I regularly work in environments below 10 lux.

Well, 10 is only for the **threshold**, so that would be fine, I guess?

> Looks like you have kept the logarithmic scale though, right?

I sure did! 10-1000 for both the "slider" and the NumberBoxes. The ticks and labels are placed dynamically and easy to change.

> Please do not push yet as I am working on more stuff. Hoping for us to find a way to merge these two approaches.

Okay, I hope we find a way then...

> Btw just noticed it says "Time" as the header for the section which obviously needs to change now 😅

Yep, I thought of that too... Has to be a different PR because that would also include file names and MVVM names etc.

BYK:
>Well, 10 is only for the threshold, so that would be fine, I guess?

No because I may actually go lower than 10 😅

Jay-o-Way:
LOL so now we have two versions of this branch 😅

BYK:
@Jay-o-Way pretty sure we can converge :)

BYK:
Added auto calibration based on the current light level (can remove if it sounds silly or not very useful)

Jay-o-Way:
@BYK here are the files that are changed in my version. I am very proud to get it working ánd looking so nice. Hope we can keep the best of both!
[Changes to PR from Jay.zip](https://github.com/user-attachments/files/24625411/Changes.to.PR.from.Jay.zip)

Update: minimum = 1 lux
image

BYK:
@Jay-o-Way I'll do my best to incorporate them into my current state - thanks a lot for sharing!

Armin2208:
Thank you very much for contributing to the Auto Dark Mode project! Your implementation of this feature looks promising and beautiful. I will test it on my Microsoft Surface Pro 7, which I want to sell, but I hold the device currently for some ambient sensor testing. 

You really pulled some work off me. Because I planned implementing this feature this year, but I don't have the time and would need to sacrifice other areas of my life. So really thank you thank you thank you! This is a highly requested feature and could tick off one box on #15.

I originally planned to implement this as an additional condition, but maybe its easier and more intelligent to implement this as a theme switch type as replacement of time. What do you think @Spiritreader?

**Some suggestions and thoughts:**

1) This feature definitely needs a second timer, because switching immediately after the threshold was reached could result in random, unwanted switches. So it probably needs to configurable second timer (for example wait 15 seconds before switching).

2) If a device doesn't have a light sensor, this probably shouldn't be a viewable setting, just like enable dark mode while on battery. 

3) because the code is AI written, we need to check the integrity carefully.

Jay-o-Way:
> 1) This feature definitely needs a second timer, because switching immediately after the threshold was reached could result in random, unwanted switches. So it probably needs to configurable second timer (for example wait 15 seconds before switching).

Agreed.

> 2) If a device doesn't have a light sensor, this probably shouldn't be a viewable setting, just like enable dark mode while on battery. 

True. I tried the **Visibility** property, that does hide the control, but it reserves the space it would use. Using `x:Load=` doesn't build. So, until we find a better way, **IsEnabled** is okay.

> 3) because the code is AI written, we need to check the integrity carefully.

Still suspicious eh?

Spiritreader:
Hi! Backend guy chiming in!

First of all thanks @BYK for the massive effort you went through to implement this feature! The hooks into the Governor system all look fine as well as how you modified the ThemeManager. 

(At some point we plan on making registering hooks in the ThemeManager more streamlined with some form dependency injection, but for now this is the intended way).

A couple of notes I had while reviewing:

- You've opted not to hook the Governor into the Auto Switch Notification system, which I consider to be a correct decision (I mean, who wants to be notified every time). If possible, please add a comment there stating that the hook is not necessary due to User Experience.
- You've renamed TimeSource to TriggerMode, useful and well understandable

I like that you've kept the implementation where minimal. I can't test edge cases as I don't have an Ambient Light sensor, but from my point of view this all looks very good.

In terms of frontend, I'm leaving that in the capable hands of @Jay-o-Way , @ChenYiLins and @Armin2208 who I think have given fantastic feedback on this. I'm going to approve the backend changes.

@AutoDarkMode/ui please let me know when you consider this feature merge ready, then I'll briefly do a sanity check and merge.

@BYK Could you change the target branch to the v11.1 branch please?
That's the current feature branch for the upcoming version and is also the one the beta targets.

Spiritreader:
> Wow, I! ! This is a very great PR! ! ! Thank you very much for your contribution! ! ♪(^∇^*) Although I noticed that you said that this PR is assisted by AI, I believe that as long as there are enough tests, this will not be a problem!

@ChenYiLins Yeah, the backend code appears to be solid. I verified that there are only changes where they should be, and from what I can tell it looks in order.

Armin2208:
Hey, I tested this feature on my Surface Pro 7. It works great and looks nice. It's easy to configure, because the UI is decent.

Some suggestions: 

1) Auto Dark Mode should recall the saved values, which the user set for the light and dark threshold, even after switching to a different Time mode. Currently, the values always get reset to the default ones (0 and 2).

2) the Calibrate feature is a little bit weird, because it sets the _Dark below_ value way too high. This feature needs more calibration, but improvements to that should rely on some more advanced math.

3) The description of the feature on the Time page needs an additional sentence at the beginning, saying: _This feature automatically switches the theme based on the brightness of your environment_. For better clarity for the users.

Beside of that, and the feedback I gave before with the time threshold option, I think the feature works pretty well and is really satisfying to use!

ChenYiLins:
The backend changes are perfect! Clean and tidy! I have added a little code for the time threshold option mentioned by Armin, and the default switching waiting time is 15 seconds (I don't know if this time is too long, or we can add an option in the UI so that users can adjust it by themselves).

@BYK Because I don't have test equipment, I can't guarantee whether the code can run in an ideal state. If you think there is something inappropriate, please modify it at will! ( •̀ ω •́ )✧

ChenYiLins:
> Hey @ChenYiLins if I read correct, the addition you made is a delay _before_ applying, right? (I think) what is wanted is a cooldown period _after_ a switch. And yes, 15 seconds is also a random number. But at least it is customizable 👍🏻

I don't think some problems can be avoided with the delay after switching (such as accidentally blocking the ambient light sensor, which will still lead to switching because of the delay after switching).

As Armin also mentioned, before the swithcing.

> So it probably needs to configurable second timer (for example wait 15 seconds before switching).

BYK:
Hi everyone and thanks so much for the support!

>So it probably needs to configurable second timer (for example wait 15 seconds before switching).

I've been using this modified build since the PR is up and I don't think we need the second timer since we have 2 different tresholds which does hysteresis. So if we're gonna add a second timer, then we should probably have a single treshold (which I'm also OK with). That said I feel a second timer would be more confusing from a UX perspective.

>Auto Dark Mode should recall the saved values, which the user set for the light and dark threshold, even after switching to a different Time mode. Currently, the values always get reset to the default ones (0 and 2).

I'll look into this, good point! There's also the new "Calibrate" button which takes your current situation as the "light normal". That said if you think this is not very useful, I'm OK removing that.

>the Calibrate feature is a little bit weird, because it sets the Dark below value way too high. This feature needs more calibration, but improvements to that should rely on some more advanced math.

Yeah, I can take away as mentioned. It just does a 15% spread around the current value I'd say and doesn't do any real calibration. A real calibration would have watched your values over a period and observe if you toggle themes etc. Not sure if that would be worth the complexity tho.

>The description of the feature on the Time page needs an additional sentence at the beginning, saying: This feature automatically switches the theme based on the brightness of your environment. For better clarity for the users.

Fully agree. I'd say it should no longer say "Time" but "Auto switch" or "Automation" with a change of icon. Thoughts?

BYK:
>You've opted not to hook the Governor into the Auto Switch Notification system, which I consider to be a correct decision (I mean, who wants to be notified every time). If possible, please add a comment there stating that the hook is not necessary due to User Experience.

This was all Opus 4.5 so cannot take much credit for that. The name change was me though :D

Jay-o-Way:
@BYK 
- I get that a timer can feel too heavy when there is already a "gray area" between the two thresholds. Up to you.
- Renaming the "time" headers/etc.is good, but can be done in a new pr to keep things overseeable.
- Is there anything else you are working on?

Armin2208:
> The backend changes are perfect! Clean and tidy! I have added a little code for the time threshold option mentioned by Armin, and the default switching waiting time is 15 seconds (I don't know if this time is too long, or we can add an option in the UI so that users can adjust it by themselves).

The reason behind the waiting time is to encounter abrupt light changes. For example on tablets, if the user gets with the hand over the light sensor (because he wants to tap something).
15 seconds is actually a good default value.

And yes, it needs to be a timer before switching, to encounter these problems. 

> I've been using this modified build since the PR is up and I don't think we need the second timer since we have 2 different tresholds which does hysteresis. So if we're gonna add a second timer, then we should probably have a single treshold (which I'm also OK with). That said I feel a second timer would be more confusing from a UX perspective.

I am also okay with just one toggle, which says like "delayed sensor readout for less discruptions" or something like that. 
or "smoother transition", not explaining any technical reasons.

**Alternatively, we can launch the feature WITHOUT A TIME SELECTION AT ALL. Because 15 seconds, no one will ever notice or complain about.**

Because only giving an option, like 15 seconds, would actually work. The user probably don't needs a choice (and if he needs it, we will get some issues on GitHub).
It depends on your work environment, if you need these seconds. I think no one would complain about a 15 seconds delay.
But as I said: as a tablet user you will encounter way more accidental switches compared to a laptop user.
And if you work outside, you will encounter more random switches compared to insight your home/office.

> I'll look into this, good point! There's also the new "Calibrate" button which takes your current situation as the "light normal". That said if you think this is not very useful, I'm OK removing that.

Generally having a calibrate button would be insanely nice, so lets keep it and improve it.

> Yeah, I can take away as mentioned. It just does a 15% spread around the current value I'd say and doesn't do any real calibration. A real calibration would have watched your values over a period and observe if you toggle themes etc. Not sure if that would be worth the complexity tho.

EDIT: I tested it again. The first time pressing calibrate results in good values. But you can always press it a second time and kinda smash the values.
The dark mode value gets way too high, and the light value gets even higher.

> Fully agree. I'd say it should no longer say "Time" but "Auto switch" or "Automation" with a change of icon. Thoughts?

We probably can't rename the whole page again currently, this needs more thinking. But I know what you mean, I will track that.
"Automation" could be a rename for this page and the titles in the future. Need to decide that further and separately.

Armin2208:
I made a little video about my feedback, to make the issues better understandable:
https://youtu.be/lvtkGWwvfu4

nuget.config
Armin2208, line 1: we removed this file from the repo on the newest version, because it results in building errors. It's also now in our .gitignore file.

BYK:
Okay, here are the latest changes:

1. Rebased onto v11.1 and fixed conflicts
1. Incorporated @Jay-o-Way's earlier changes
1. Added a 10 second delay between switches (tested with 15, felt too long)
1. Improved calibration:
    - No longer does random stuff on multiple clicks
    - Takes the current theme and lighting situation into account
    - Has a minimum for 10 lux
1. Added a smooth animation for the current lighting indicator
1. Added a disabled mode along with a tooltip when a light sensor is not found

BYK:
@Jay-o-Way, @Armin2208 -- I think this is ready for another spin

AutoDarkModeApp/Views/TimePage.xaml
```diff
@@ -202,7 +242,326 @@
...
+                                <Grid HorizontalAlignment="Stretch" ColumnSpacing="16">
+                                    <Grid.ColumnDefinitions>
+                                        <ColumnDefinition Width="Auto" />
+                                        <ColumnDefinition Width="Auto" />
+                                        <ColumnDefinition Width="*" />
+                                        <ColumnDefinition Width="Auto" />
+                                        <ColumnDefinition Width="Auto" />
+                                    </Grid.ColumnDefinitions>
+                                    <Grid.RowDefinitions>
+                                        <RowDefinition Height="Auto" />
```
Jay-o-Way, line 412: Note that this row (`Grid.Row="0"`) isn't actually used :)

AutoDarkModeApp/Views/TimePage.xaml
```diff
@@ -202,7 +242,326 @@
...
+                                        SizeChanged="IndicatorCanvas_SizeChanged">
+                                        <FontIcon
+                                            x:Name="CurrentLuxIndicator"
+                                            FontSize="12"
+                                            Foreground="{ThemeResource AccentTextFillColorPrimaryBrush}"
+                                            Glyph="&#xE96E;" />
+                                    </Canvas>
+
+                                    <!--  Dark threshold NumberBox  -->
+                                    <NumberBox
```
Jay-o-Way, line 418: FYI: setting **Width** may wrap the description (if text is longer)

AutoDarkModeApp/Views/TimePage.xaml
```diff
@@ -202,7 +242,326 @@
...
+                                            Foreground="Transparent"
+                                            Loaded="AmbientLightRangeSelector_Loaded"
+                                            Maximum="1000"
+                                            Minimum="0"
+                                            RangeEnd="{x:Bind ViewModel.RangeEnd, Mode=TwoWay}"
+                                            RangeStart="{x:Bind ViewModel.RangeStart, Mode=TwoWay}"
+                                            StepFrequency="1" />
+                                    </Grid>
+
+                                    <!--  Tick marks with lux reference labels (logarithmic positions)  -->
```
Jay-o-Way, line 492: @BYK What did you think about the other solution, with the tickmarks being made and placed all in code-behind, inside a Canvas?

AutoDarkModeApp/Views/TimePage.xaml
```diff
@@ -202,7 +242,326 @@
...
+                                                Width="1"
+                                                Height="6"
+                                                HorizontalAlignment="Center"
+                                                Background="{ThemeResource TextFillColorTertiaryBrush}" />
+                                            <TextBlock
+                                                HorizontalAlignment="Center"
+                                                FontSize="10"
+                                                Foreground="{ThemeResource TextFillColorTertiaryBrush}"
+                                                Text="10" />
+                                        </StackPanel>
```
Jay-o-Way: This is so much XAML for each tickmark... _If_ we choose to keep this approach, these things need a Style.

AutoDarkModeApp/Views/TimePage.xaml.cs
```diff
@@ -12,6 +15,53 @@ public TimePage()
...
+    {
+        UpdateIndicatorPosition();
+    }
+
+    private void UpdateIndicatorPosition()
+    {
+        if (IndicatorCanvas == null || CurrentLuxIndicator == null) return;
+
+        double width = IndicatorCanvas.ActualWidth;
+        if (width <= 0) return;
```
Jay-o-Way, line 40: This should be absolutely impossible to get?

AutoDarkModeApp/ViewModels/TimeViewModel.cs
```diff
@@ -91,12 +94,160 @@ public enum TimeSourceMode
...
+        get => _ambientLightDarkThreshold;
+        set
+        {
+            value = Math.Round(value);
+            if (SetProperty(ref _ambientLightDarkThreshold, value))
+            {
+                if (!_isUpdating)
+                {
+                    _isUpdating = true;
+                    // Ensure Light stays above Dark
```
Jay-o-Way: In the control, the left thumb can not _pass over_ the other thumb to become the right thumb. But they can be the _same_ value.

ChenYiLins:
Here I found a few small problems:

1. Should Inline be used for the SpinButtonPlacementMode property of NumberBox? When using the Compact mode, sliding the RangeSelector will cause obvious jamming when the SpinButton is expanded, although it will occupy more UI space.

PixPin_2026-01-20_11-45-03

2. We should not use the slider as a precise operation. On the contrary, the NumberBox should be used as a precise operation.

AutoDarkModeApp/ViewModels/TimeViewModel.cs
```diff
@@ -222,31 +471,38 @@ private void HandleAutoTheme(bool value)
...
+            SelectedTriggerMode = SwitchTriggerMode.CustomTimes;
             TimePickerVisibility = Visibility.Visible;
             DividerBorderVisibility = Visibility.Collapsed;
             return;
         }
 
         if (_builder.Config.Location.UseGeolocatorService)
         {
-            SelectedTimeSource = TimeSourceMode.LocationTimes;
-            OffsetTimeSettingsCardVisibility = Visibility.Visible;
```
ChenYiLins, line 244: This should be an accidental deletion? The Offset function should be enabled.

AutoDarkModeApp/Views/TimePage.xaml.cs
BYK: This is a defensive check for edge cases during layout passes. ActualWidth can be 0 before the layout system completes measuring, especially when the page is first loading or during certain resize scenarios. While rare in practice, it prevents potential divide-by-zero or invalid positioning calculations.

AutoDarkModeApp/Views/TimePage.xaml
BYK: I considered the code-behind Canvas approach. For this case, I opted to keep the tickmarks in XAML because: (1) the tick positions are static/fixed based on logarithmic scale, (2) it keeps the visual layout declarative and easier to adjust in XAML designer, and (3) the number of ticks is small (5 total). I have added styles to reduce the repetition as suggested in the follow-up comment.

AutoDarkModeApp/Views/TimePage.xaml
BYK: Good point! I have added TickMarkStackPanelStyle, TickMarkBorderStyle, and TickMarkTextBlockStyle to Page.Resources and refactored all tickmarks to use them. Much cleaner now.

AutoDarkModeApp/Views/TimePage.xaml
BYK: Acknowledged. The fixed width is intentional to maintain consistent alignment between the Dark and Light threshold NumberBoxes. The header text (which is localized) is kept short to fit within this width. If longer translations cause issues, we can revisit.

AutoDarkModeApp/Views/TimePage.xaml
BYK: Thanks for catching this! This Grid.Row attribute was already removed in a subsequent commit - the reviewed diff might have been from an older version.

AutoDarkModeApp/ViewModels/TimeViewModel.cs
BYK: Good point! I intentionally enforce Light > Dark (not >=) to prevent an ambiguous zone where both thresholds are equal. If they were the same value, the system would not know whether to apply dark or light theme at that exact lux level. The minimum gap of 1 lux provides clear hysteresis for theme switching. I have added a detailed comment explaining this design decision in the code.

AutoDarkModeApp/ViewModels/TimeViewModel.cs
BYK: You are absolutely right - this was an accidental deletion. I have restored OffsetTimeSettingsCardVisibility = Visibility.Visible for LocationTimes/CoordinateTimes modes. Thank you for catching this!

BYK: >You are absolutely right 

You can tell this was the LLM 🤣

BYK:
@ChenYiLins changed based on your input but honestly, not a fan of this new look. Looks too crowded to me:
image

ChenYiLins:
> @ChenYiLins changed based on your input but honestly, not a fan of this new look. Looks too crowded to me

Yes, indeed. Let me see if there is any way to solve this problem from the code level. (っ °Д °;)っ

BYK:
@ChenYiLins went with removing the spinners altogether and relying on keyboard up/down arrow buttons. We have the sliders for the spinner action anyway.

Armin2208:
Hello, I checked the newest commit version of this fork. I did use my Surface way more than in the last months lol.

1)  I prefer the information icon being right next to the text "Ambient light sensor". I think having it at the right is kinda weird. Additionally, if the feature is disabled, the information text should be extended by one more sentence, saying: "This feature isn't available on your device, because we didn't recognize an Ambient Light Sensor".
grafik

2) The schedule of the switch is 10 minutes, but in the configuration file the DebounceDelayMs is still set on 15000 (default value).
So please also update the default value on the config file and make the option in the config file workable again. Yes, there will be no UI option for the delay. But at least make it configurable in our config file. In general, the config file of Auto Dark Mode supports more option compared to the User Interface.

3) Your changes on the calibration features are pretty good. The light above value is now always understandable. And multiple times clicking it doesn't break it anymore. Great.
On my surface, with 41 lux room light, the values are set like this: dark below 28, light above 37. The gap is not so high, maybe the logik could make the gaps a little bit higher (for smaller numbers? 

4) Maybe it's not completely clear, that the left lux number is the current sensor readout. Maybe we could name it like that? The space is not big though.

5) I don't know what happened to the slider, but I can't slide it with touch anymore. Only tapping works now.

Fixing these five issues should make this feature Pull Request ready.

BYK:
@Armin2208 Thanks for the detailed feedback! Here's a summary of the changes made:

**Addressing your review comments:**

1. **Info icon position & disabled tooltip** - Moved the info icon inline with "Ambient light sensor" text. Added a conditional tooltip that shows "This feature isn't available on your device because no ambient light sensor was detected" when no sensor is present.

2. **Debounce delay configuration** - Fixed the mismatch: updated config default from 15000ms to 10000ms, and changed the `Math.Max(15000, ...)` calls to use the `SensorDebounceDelayMs` constant so configured values below the hardcoded limit are now respected.

3. **Current lux display** - Added a "Current:" label before the lux reading for clarity.

4. **Slider touch issue** - Added `IsHitTestVisible="False"` to the IndicatorCanvas so touch events pass through to the RangeSelector.

**Calibration gap logic (your latest comment):**

You noted that at 41 lux, the gap (28-37 = 9 lux) was too narrow. We've changed the calibration formula from a fixed 1:1.3 ratio to an exponential scaling approach:

```csharp
double gap = Math.Pow(currentLux, 0.7);
```

This creates non-linear scaling where:
- 10 lux -> ~5 lux gap
- 41 lux -> ~13 lux gap (was 9)
- 100 lux -> ~25 lux gap
- 1000 lux -> ~126 lux gap

The gap grows proportionally larger at higher lux values, which better accounts for sensor noise at brighter conditions while keeping tight control at low light levels.

Armin2208:
Thanks for the fast and good implementation of our feebdack!
Something went wrong while moving the information button, as you can see on the screenshot.
grafik

Armin2208:
I changed the milisecond time in the config file to 30000 but it still switches after 10 seconds and the log file also says: switching in 10 seconds.
Seems like it doesn't read the config file correctly / or it doesn't work with the config value correctly.

Armin2208:
It seems like the Ambient Light Sensor option on the time page still invokes a location access request, because every launch of the page and every modifications of the slider creates following log file:
"MessageParser.Parse: signal received: checking location access permissions"

Armin2208:
touch operations still doesn't work, even with the new change trying to fix that. It seems like this issue has a different origin. Maybe I can test changing some values on my Surface (but I need to install Visual Studio for that 😅)

ChenYiLins:
> Thanks for the fast and good implementation of our feebdack! Something went wrong while moving the information button, as you can see on the screenshot.

Fixed it.

image

Armin2208:
Hey! I have again a bunch of feedback:

1) Touch input still doesn't work.
2) If I disable "Automatically switch themes" (the first toggle of the time page), the AmbientLightGovernor still runs and changes the theme.
3) if threshold gets modified in the config file (not in the UI), there will be no theme switch.
4) you can ask postpone manager if an active postpone delay is enabled, to stop logging while postpone delay is active.

Sorry for coming out with even more things, but because the groundwork works now pretty good, I checked some more advanced integrity stuff.
But these 4 issues should be the last ones, after that this is totally merge ready and totally fine.

Spiritreader:
> Hey! I have again a bunch of feedback:
> 
>     1. Touch input still doesn't work.
> 
>     2. If I disable "Automatically switch themes" (the first toggle of the time page), the AmbientLightGovernor still runs and changes the theme.
> 
>     3. if threshold gets modified in the config file (not in the UI), there will be no theme switch.
> 
>     4. you can ask postpone manager if an active postpone delay is enabled, to stop logging while postpone delay is active.
> 
> 
> Sorry for coming out with even more things, but because the groundwork works now pretty good, I checked some more advanced integrity stuff. But these 4 issues should be the last ones, after that this is totally merge ready and totally fine.

Yes, the postpone manager is registered and available in the GlobalState singleton an exposes the `IsPostponed` property to be queried: https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/bc6e49461f343dd7ae1e00c6b5d3a02513b63245/AutoDarkModeSvc/Core/PostponeManager.cs#L93

Also, as we're getting close to getting this merged, I've incorporated the latest changes into the current beta, and this PR is first in line on being pushed out to users on the 11.1 beta as well once it gets merged.

BYK:
@Armin2208 no worries, pushed some new changes

@Spiritreader - thank you!

Armin2208:
I tested:
1) DelayTimer needs to be at least 10 seconds, I would like users to be able to put lower values than 10 seconds, if they want.
2) touch still doesn't work, but the newest changes results in the green line of the slider not showing. I don't know if this is the real cause of this touch issue.

Spiritreader:
Something we also might need to consider.

ADM has feature to toggle the theme from the tray or a hotkey.
For the Default Time- and NiightLight Governor, this means that the automatic switching is paused untl the next day/night cycle (if it's night and you toggle to light theme, it will stay that way until the next day when it gets dark again)

For the ambient light mode, this doesn't really make sense as we have no concept of day or night.

https://github.com/BYK/Windows-Auto-Night-Mode/blob/d6f6e2634fff04479acb4e6f79f6c8bdb86f4d5b/AutoDarkModeSvc/Core/ThemeManager.cs#L148

This is the function where we would add such a behavior.
Currently, if no specific option is implemented here, it will just toggle the theme.

Example: Room is dark, user selects "toggle theme", theme switches to Light mode.
Then the room gets bright, no action is done, the room gets dark again, Auto Dark Mode switches back.
If we *don't* want this behavior, we need to define what the pausing logic for the "toggle theme" feature should be.
A time delay perhaps? (can queue with the postpone system)

Another thing to consider is the "pause theme switch" feature.
Because with the ambient sensor mode, it's not different from enabling/disabling the ADM auto switching feature altogether. The implementation also requires a postpone element queued for this to work, as such we should probably hide the menu item. Or if we choose to queue a time delay with the postpone system, make it clear that of course.

Armin2208:
Great work, this Pull Request passed all checks - it works well, its integrated and polished. 

We still can't merge it, becaues on thing isn't resolved: the touch input. This issue probably has a different origin. But the changes in the recent builds made the arrow being far more away from the nice looking line before. So I would recommend reverting these changes or optimizing the UI spacing, and always double check if the UI behaves and looks as expected.

And the accent color line between threshold light and dark is still missing.

If you have an idea which xaml values I should modify or test on my Surface, let me know.

Screenshot 2026-01-22 135102

Armin2208:
or if @ChenYiLins and @Jay-o-Way have an idea, why touch doesn't work / also have a touch device to test, also let us know. Because this issue is the last remaining before the merge and the origin seems unkown yet.

BYK:
Thanks for all the help and support @Armin2208 

**Changes made:**
1. **Removed `Foreground="Transparent"` from RangeSelector** - This was hiding the accent-colored selected range indicator (the "green line" between thresholds). Now the default accent color will show.
2. **Reverted IndicatorCanvas to original working layout** - The touch issue was introduced when we started moving the Canvas around. I traced back to commit 92e048c1 where touch was working and restored that layout:
  - Moved Canvas back to `Grid.Row="1"` (same cell as RangeSelector)
  - Removed the `Height="20"` constraint
  - Restored original `Margin="10,12,10,0"` and `VerticalAlignment="Top"`
**Why this works:** The Canvas and RangeSelector share the same grid cell, but different margins position the Canvas visually above the slider track. Since the Canvas is declared before RangeSelector in XAML, it renders behind (lower z-order), so touches go through to the RangeSelector. The `IsHitTestVisible="False"` on Canvas also ensures it doesn't intercept input.

This report was generated by the LLM again and as usual, I have very little understanding of what any of this means 🤣

Jay-o-Way:
UX looks nice now. Just one little thing: When I lower the value of the left **numberbox** down to 1, the slider thumb does not go to the left end. This must be because of rounding the value. When I move the **thumb** itself from that place to the actual end, the value stays 1.
image

Armin2208:
So, my newest test showed following:

1. It's still not possible to **move** the slider with touch, you can just **tap** it and it will change the slider position. [edit: fixed]
2. While Force Light Mode and Force Dark Mode is activated, AmbientLightGovernor still writes in the log. Just like pause theme switching, this should be removed. (Yeah, that's kinda complex. Note that we want to remove force switching in the future, but currently we don't have the time to do that, so the AmbientLightGovernor should react to that). @BYK @Spiritreader 
3. With the AmbientLightGovernor you can't use the postpone option "Skip auto switch one". We should hide that on the App interface on the combobox @ChenYiLins.
4. In the tray menu, it should say "Pause auto switch (for one hour)" if the AmbientLightGovernor is active @ChenYiLins @Spiritreader.
grafik

Jay-o-Way:
> * It's still not possible to **move** the slider with touch, you can just **tap** it and it will change the slider position.

@Armin2208 the issue is with the **ScrollViewer**. The [Toolkit Gallery](https://aka.ms/windowstoolkitapp) specifically mentions this:

image

Note that the issue only occurs when the scrollbars are active.

> [!NOTE]
> Fixed it :)

AutoDarkModeApp/Views/TimePage.xaml
```diff
@@ -39,15 +58,53 @@
                 </InfoBar>
 
                 <!--  About time  -->
-                <controls:SettingsCard IsEnabled="{x:Bind ViewModel.AutoThemeSwitchingEnabled, Mode=OneWay}">
+                <controls:SettingsCard ContentAlignment="Vertical" Visibility="{x:Bind ViewModel.AutoThemeSwitchingEnabled, Mode=OneWay}">
```
Jay-o-Way, line 42: Wait, why was this changed to Visibility?

Jay-o-Way:
Oh, by the way, I found my old Surface Pro 4 and it works lovely on there! 👌🏻
I think the 10 second delay will be interesting for first-time users...

Armin2208:
Thank you very much for fixing this long lasting issue @Jay-o-Way. And sorry for not mentioning you.
And cool that you have a Surface Pro 4 with which you can test the feature :)
Actually, I am curious which comment you deleted from BYK. Probably the comment wasn't relevant anymore, but I still don't like comments getting deleted (and because GitHub mentions it, I am now curious).

Finally, every important issue was fixed from this PR.
Please update this Pull Request, resolve the conflicts, make it merge ready and I will approve a merge.

_I wouldn't be mad if you also fix the latest issues (especially number 2)._

Thank you
 
>2. While Force Light Mode and Force Dark Mode is activated, AmbientLightGovernor still writes in the log. Just like pause theme switching, this should be removed. (Yeah, that's kinda complex. Note that we want to remove force switching in the future, but currently we don't have the time to do that, so the AmbientLightGovernor should react to that). @BYK @Spiritreader
> 
>3. With the AmbientLightGovernor you can't use the postpone option "Skip auto switch one". We should hide that on the App interface on the combobox @ChenYiLins.
> 
>4. In the tray menu, it should say "Pause auto switch (for one hour)" if the AmbientLightGovernor is active @ChenYiLins @Spiritreader.
> 
> grafik

Jay-o-Way:
> Thank you very much for fixing this long lasting issue @Jay-o-Way. And sorry for not mentioning you.
> And cool that you have a Surface Pro 4 with which you can test the feature :)

You're welcome! Yeah i had to search for that thing haha. By the way, you commented that you would need to install Visual Studio to test? But you can also do a Rebuild and then copy the `bin\adm-app` folder to another device.

> Actually, I am curious which comment you deleted from BYK. Probably the comment wasn't relevant anymore, but I still don't like comments getting deleted (and because GitHub mentions it, I am now curious).

Oh yeah I understand. It was a small and outdated comment. Next time i wanna clean up the chat i'll just hide it.

> Finally, every important issue was fixed from this PR.
> Please update this Pull Request, resolve the conflicts, make it merge ready and I will approve a merge.

Is it an idea to get this PR in and open another for the other points? Oh and how about a _Squash_ and merge for this one? 💡

Armin2208:
> You're welcome! Yeah i had to search for that thing haha. By the way, you commented that you would need to install Visual Studio to test? But you can also do a Rebuild and then copy the `bin\adm-app` folder to another device.

I know. I actually did that all the time. But if no one had found a solution to the touch issue and I need to fix that, I would install Visual Studio on my Surface, because all the time transferring the build files takes time. Thankfully, this issue is resolved and its not necessary for me to install Visual Studio on that slow tablet.

> Is it an idea to get this PR in and open another for the other points? Oh and how about a _Squash_ and merge for this one? 💡

Possible. Actually, number 4 probably needs to be a new issue. @Spiritreader is currently ill, so he can't work on that Pull Request. But before merging this, I would prefer 2 to be fixed.

Just let me know how the status of this PR is.

Jay-o-Way:
@BYK "changes.zip" is still in the list of files

ChenYiLins:
~~Is there something wrong with the localization file(`en-us/Resources.resw`)? It can't be compiled now.~~

Fix it now.

Jay-o-Way:
> Is there something wrong with the localization file(`en-us/Resources.resw`)? It can't be compiled now.

What is the error message? Could be an issue after the Rebase... 🙄

Spiritreader:
@Jay-o-Way squash merge is fine by me in this case. I don't want to do it everywhere because it messes up git blame and reduces transparency.

But this is a single feature that gets merged as the first release candidate.

@Armin2208 and I are in contact and once he's good to go we can merge.

In terms of improving the tray wording, yes that should be a new featurea and not part of this PR.

________________

Please take care of removing the changes.zip by then. Having a random archive in the repository is odd.

ChenYiLins:
> Please take care of removing the changes.zip by then. Having a random archive in the repository is odd.

Done.

Armin2208:
I will open new Issues about this feature, feel free to improve

Jay-o-Way:
@BYK why was this changed from IsEnabled to Visibility?
```xaml
 <controls:SettingsCard ContentAlignment="Vertical" Visibility="{x:Bind ViewModel.AutoThemeSwitchingEnabled, Mode=OneWay}">
```

BYK:
@Jay-o-Way I think we wanted to hide it when there was no light sensor?

Jay-o-Way:
> @Jay-o-Way I think we wanted to hide it when there was no light sensor?

This hides the settings card - all the options, not the one

PR #1176, opened 2026-01-18
Delete nuget.config

Jay-o-Way:
@Armin2208 and I had a few issues when building the project. Deleting this file makes problems go away. 🤓


Armin2208 (approved):
also need to update gitignore, will do

Issue #913, opened 2025-02-05
Cursor theme not working


JeremyD357:
### Description

Hello, I use Auto Dark Mode 10.4.2.29, and I like to use cursor theme while it change from light to dark mode, and somehow it doesn't work. I can't explain why, I tried to uninstall and remove every folder with cleaning registry everything and still doesn't work.

### Log Data

[service.log](https://github.com/user-attachments/files/18672577/service.log)

### Commit Hash, Version and Windows Build

 - Commit: `5922db4`
- Service/App: `10.4.2.29`
- Updater: `3.1.6`
- Shell: `1.3.3.0`
- .Net: `8.0.0`
- Windows: `26100.3037`
- Arch: `X64`

JeremyD357:
For more info my config file look like this.

AutoThemeSwitchingEnabled: true
Governor: Default
Autostart:
  Validate: true
WindowsThemeMode:
  Enabled: false
  DarkThemePath: 
  LightThemePath: 
  MonitorActiveTheme: false
  ApplyFlags: []
AppsSwitch:
  Enabled: true
  Component:
    Mode: Switch
SystemSwitch:
  Enabled: true
  Component:
    Mode: Switch
    TaskbarSwitchDelay: 1200
    TaskbarColorOnAdaptive: true
    TaskbarColorWhenNonAdaptive: Light
    DWMPrevalenceSwitch: true
    DWMPrevalenceEnableTheme: Dark
TouchKeyboardSwitch:
  Enabled: false
  Component: {}
ColorizationSwitch:
  Enabled: false
  Component:
    LightHex: '#FFA1A1A1'
    DarkHex: '#FFA1A1A1'
    LightAutoColorization: false
    DarkAutoColorization: false
ColorFilterSwitch:
  Enabled: false
  Component: {}
OfficeSwitch:
  Enabled: true
  Component:
    Mode: Switch
    LightTheme: 0
    DarkTheme: 4
**CursorSwitch:
  Enabled: true
  Component:
    CursorsLight: Material Design Pure Dark v2 by Jepri Creations
    CursorsDark: Material Design Light v2 by Jepri Creations**
Sunrise: 2025-02-05T07:00:00.0000000
Sunset: 2025-02-05T20:00:00.0000000
Location:
  PollingCooldownTimeSpan: 1.00:00:00
  Enabled: true
  UseGeolocatorService: false
  SunsetOffsetMin: 0
[...]

PR #1180, opened 2026-01-20
Code quality: Check whether the registry allows theme switching cursor

ChenYiLins:
### Description

Check whether the registry allows theme switching cursor.
Sloved: #913


Spiritreader:
Great find @ChenYiLins!

Thanks for investigating this and providing a fix!

PR #1183, opened 2026-01-23
Refactor About page

Jay-o-Way:
I thought it would be nice to have a short description about the contributions, but you may edit this at will.
Closes #1178

image


Armin2208:
Thank you very much for these changes.
@Spiritreader are these edits to the localization files okay without breaking the translation platform?

Jay-o-Way:
Updated to not change all the Res files, only the English one.

Armin2208:
> Updated to not change all the Res files, only the English one.

But you did introduced new variables. How should  this work? What happens if you display the app in a different language? will the translation platform recognize that?

Jay-o-Way:
> > Updated to not change all the Res files, only the English one.
> 
> But you did introduced new variables. How should this work? What happens if you display the app in a different language? will the translation platform recognize that?

It falls back to English. ✅

image

PR #1199, opened 2026-03-07
Convert to SLNX

### LINKED ISSUE #1198: Convert SLN to SLNX

Jay-o-Way:
Closes #1198


Jay-o-Way:
I was thinking 🤔 should we even support x86?

Armin2208:
> I was thinking 🤔 should we even support x86?

No

ChenYiLins:
LGTM, the new solution file is much better than the original one! ( •̀ ω •́ )✧

PR #1202, opened 2026-03-11
Code quality: Add services for window closing

ChenYiLins:
### Description

- Instead of the original MainWindow's Close event, a new **CloseService** is created to provide the closing service corresponding to **ActivationService**, so that `MainWindow.xaml.cs` will retain the specificity of its own window event and need not care about the closing event of the whole App.
- To improve code uniformity, the **MainWindow** was originally outside the DI container, and its life cycle could not be managed by the container.
- Use DI dependency injection for the window.

### DIFF
--- a/AutoDarkModeApp/App.xaml.cs
+++ b/AutoDarkModeApp/App.xaml.cs
@@ -25,6 +25,8 @@ public partial class App : Application
     // https://docs.microsoft.com/dotnet/core/extensions/logging
     public IHost Host { get; }
 
+    // TODO: Avoid static GetService<T>(); prefer constructor injection.
+
     public static T GetService<T>()
         where T : class
     {
@@ -36,32 +38,21 @@ public partial class App : Application
         return service;
     }
 
-    public static void CheckAppMutex()
-    {
-        if (!Mutex.WaitOne(TimeSpan.FromMilliseconds(50), false) && !Debugger.IsAttached)
-        {
-            List<Process> processes = [.. Process.GetProcessesByName("AutoDarkModeApp")];
-            if (processes.Count > 0)
-            {
-                Helpers.WindowHelper.BringProcessToFront(processes[0]);
-                App.Current.Exit();
-            }
-        }
-    }
-
-    public static Window? MainWindow { get; set; }
+    // Expose the DI-created MainWindow for legacy call sites.
+    // This preserves compatibility with existing App.MainWindow usages.
+    // Consider refactoring callers to use dependency injection directly.
+    public static MainWindow MainWindow =>
+        (Current as App)!.Host.Services.GetRequiredService<MainWindow>();
 
     public App()
     {
         CheckAppMutex();
-
         InitializeComponent();
 
-        Host = Microsoft
-            .Extensions.Hosting.Host.CreateDefaultBuilder()
+        Host = Microsoft.Extensions.Hosting.Host
+            .CreateDefaultBuilder()
             .UseContentRoot(AppContext.BaseDirectory)
-            .ConfigureServices(
-                (context, services) =>
+            .ConfigureServices((context, services) =>
                 {
                     // Services
                     services.AddSingleton<ILocalSettingsService, LocalSettingsService>();
@@ -115,6 +106,20 @@ public partial class App : Application
         UnhandledException += App_UnhandledException;
     }
 
+    public static void CheckAppMutex()
+    {
+        if (!Mutex.WaitOne(TimeSpan.FromMilliseconds(50), false) && !Debugger.IsAttached)
+        {
+            var processes = Process.GetProcessesByName("AutoDarkModeApp").Where(p => p.Id != Environment.ProcessId).ToList();
+            if (processes.Count > 0)
+            {
+                Helpers.WindowHelper.BringProcessToFront(processes[0]);
+                App.Current.Exit();
+            }
+        }
+    }
+
+
     private void App_UnhandledException(object sender, Microsoft.UI.Xaml.UnhandledExceptionEventArgs e)
     {
         // TODO: Log and handle exceptions as appropriate.
@@ -131,21 +136,23 @@ public partial class App : Application
         {
             new PipeClient().SendMessageAndGetReply(arguments[1]);
             App.Current.Exit();
+            return;
         }
 
-        // Set App and Svc language
+        // Set App and SVC language
         Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = await LanguageHelper.GetDefaultLanguageAsync();
         await Task.Run(() =>
         {
             var builder = AdmConfigBuilder.Instance();
             builder.Load();
-            builder.Config.Tunable.UICulture = LanguageHelper.SelectedLanguageCode;
+            builder.Config.Tunable.UICulture = LanguageHelper.SelectedLanguageCode; // For SVC and other services that need to know the UI culture
             builder.Save();
         });
 
-        MainWindow = App.GetService<MainWindow>();
-        MainWindow.Closed += async (s, e) => await App.GetService<ICloseService>().CloseAsync();
+        var mainWindow = Host.Services.GetRequiredService<MainWindow>();
+        mainWindow.Closed += async (s, e) => await Host.Services.GetRequiredService<ICloseService>().CloseAsync();
 
-        await App.GetService<IActivationService>().ActivateAsync(args);
+        await Host.Services.GetRequiredService<IActivationService>().ActivateAsync(args);
+        mainWindow.Activate();
     }
 }
--- a/AutoDarkModeApp/App.xaml.cs
+++ b/AutoDarkModeApp/App.xaml.cs
@@ -67,11 +67,15 @@ public partial class App : Application
             services.AddSingleton<IFileService, FileService>();
 
             services.AddSingleton<IActivationService, ActivationService>();
+            services.AddSingleton<ICloseService, CloseService>();
             services.AddSingleton<IPageService, PageService>();
             services.AddSingleton<INavigationService, NavigationService>();
 
             services.AddSingleton<IErrorService, ErrorService>();
             services.AddSingleton<IGeolocatorService, GeolocatorService>();
+
+            // Window
+            services.AddSingleton<MainWindow>();
 
             // Views and ViewModels
             services.AddTransient<ThemePickerViewModel>();
@@ -142,7 +146,8 @@ public partial class App : Application
             //await App.GetService<IErrorService>().ShowErrorMessage(ex, App.MainWindow.Content.XamlRoot, "Startup", "Failed to force-set Svc language");
         }
 
-        MainWindow = new MainWindow();
+        MainWindow = App.GetService<MainWindow>();
+        MainWindow.Closed += async (s, e) => await App.GetService<ICloseService>().CloseAsync();
 
         await App.GetService<IActivationService>().ActivateAsync(args);
     }
--- a/AutoDarkModeApp/App.xaml.cs
+++ b/AutoDarkModeApp/App.xaml.cs
@@ -57,58 +57,60 @@ public partial class App : Application
 
         InitializeComponent();
 
-        Host = Microsoft.Extensions.Hosting.Host.
-        CreateDefaultBuilder().
-        UseContentRoot(AppContext.BaseDirectory).
-        ConfigureServices((context, services) =>
-        {
-            // Services
-            services.AddSingleton<ILocalSettingsService, LocalSettingsService>();
-            services.AddSingleton<IFileService, FileService>();
+        Host = Microsoft
+            .Extensions.Hosting.Host.CreateDefaultBuilder()
+            .UseContentRoot(AppContext.BaseDirectory)
+            .ConfigureServices(
+                (context, services) =>
+                {
+                    // Services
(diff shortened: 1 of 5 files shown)


Jay-o-Way:
I'm just struggling to understand the reason for this solution in the first place. Can you (or Armin or Sam) explain? I mean, I thought this is an ordinary Window Close event, right?

A window close event (e.g., clicking the X on the tray UI) is:

- a user‑initiated, routine action
- not an error condition
- not a recovery scenario
- not a sign that the service is hung

Using it to trigger a kill‑restart is like slamming the brakes every time you park your car. It works, but it’s not what the mechanism is for.

---

Closing a WinUI Window ≠ Closing the App

In WinUI 3 (packaged or unpackaged), a Window is just a UI surface.  
Closing it:

- disposes the window  
- ends its dispatcher loop  
- removes it from the visual tree  

…but the process keeps running as long as:

- other windows exist, or  
- background tasks are running, or  
- the app’s message loop is still active  

This is exactly what you want for AutoDarkMode:  
the UI can close while the service keeps running normally.

A clean, expected behavior:

1. User clicks the X  
2. The WinUI window closes  
3. The tray icon may remain or disappear depending on design  
4. The AutoDarkMode service continues running in the background  
5. Nothing else restarts or gets killed  

This is how almost all tray‑based Windows utilities behave.

ChenYiLins:
@Jay-o-Way I don't know if my explanation can dispel your doubts. This Kill command is only used to solve the problem that the UI thread will be suspended, and the tray icon will be kept, because the tray is handled by **Svc**, not App, and the forced termination of **App** will not have any impact on the service. For **App**, its life cycle is the life cycle of the window, and when the window is closed, the **App** will end.

Jay-o-Way:
Yes, but a window should be CLOSED, that is with a `this.Close`, not KILL the entire process. And to add to add extras, you _catch_ the Close event. 

```cs
window.Closing += (sender, args) =>
{
    // optional: cancel close
    // args.Cancel = true;

    // do cleanup
};
```

That code is not a “close handler.” It is a full application suicide wired to a normal window‑close event. No WinUI app should ever behave like this. Even with a separate tray process, killing the UI process on window close is still harmful:

❌ It bypasses WinUI cleanup
No disposal, no state flush, no graceful teardown.

❌ It can break IPC
If the UI communicates with the tray or service, killing it can leave pipes/sockets/COM channels in a bad state.

❌ It prevents the app from reusing the same process for future windows
Normally, you can reopen the settings window without restarting the whole app.

❌ It’s not how Windows apps behave
Users expect the window to close, not the process to die.

❌ It’s a debugging hack, not production logic
This pattern is almost always a leftover from trying to escape a deadlock during development.

If you want to fully exit the UI
Call:

```csharp
App.Current.Exit();
```

I know this was here long before this PR, so in fact this is directed at @Armin2208 and @Spiritreader

ChenYiLins:
@Jay-o-Way I have deleted the code of the Kill part, and also fixed the Bug that the program was hang on. On the one hand, it is related to the original `MapLocationFinder`, but now we don't use it. In principle, this bug won't appear. Finally, when I locked the original initialization language, due to the existence of static references, when the DI container tried to Dispose, the object could not be recovered normally.

Now we don't have to manually interfere with the shutdown of the program.

AutoDarkModeApp/App.xaml.cs
```diff
@@ -129,20 +135,16 @@ protected async override void OnLaunched(LaunchActivatedEventArgs args)
 
         // Set App and Svc language
         Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = await LanguageHelper.GetDefaultLanguageAsync();
-        var builder = AdmConfigBuilder.Instance();
-        builder.Load();
-        try
+        await Task.Run(() =>
```
Jay-o-Way, line 149: why is this block changed?

ChenYiLins: Because ActivationService starts to execute after the language is loaded, configuration file loading does not need to block UI threads, and it also unifies asynchronous business logic. (There won't even be performance improvement, but it looks much better visually)

AutoDarkModeApp/App.xaml.cs
```diff
@@ -44,12 +44,12 @@ public static void CheckAppMutex()
...
             {
                 Helpers.WindowHelper.BringProcessToFront(processes[0]);
-                Environment.Exit(-1);
+                App.Current.Exit();
             }
         }
     }
 
-    public static Window MainWindow { get; set; } = Window.Current;
+    public static Window? MainWindow { get; set; }
```
ChenYiLins: Is the change here unnecessary? The `Window.Current` property does always return null, but here it is actually through human intervention in the compiler's interpretation of the life cycle. Because if `MainWindow = App.GetService<MainWindow>();` fails, the program will not be executed in the next step, and now the entire application life cycle of App is bound to the window life cycle, and the **MainWindow** should never be null in business logic. The code you modified does make the program more secure and honestly reflect when the window exists and does not exist, but we need to introduce a lot of unnecessary null checks, which are unnecessary. My suggestion is that if we want to remove the ambiguous initial value of `Window.Current`, we can change it to `public static Window MainWindow { get; set; } = null!;`.

Jay-o-Way: Looks like both "versions" are possible and it's difficult to understand the difference. But advice from my anonymus co-worker says:

> `public static Window MainWindow { get; set; } = null!;`
> This communicates:
> - “This property will definitely be non-null before anyone uses it.”
> - “Trust me, the compiler warning is wrong.”
>
> But in your app, that’s not guaranteed. If any code accesses App.MainWindow before OnLaunched runs, it _will_ be null. So this declaration is misleading.

ChenYiLins: This statement seems to be no problem. I just think that adding null checking is unnecessary overhead, just like doing it in [WinUI Gallery](https://github.com/microsoft/WinUI-Gallery/blob/e52ce61f85794afa01e379d77b8c029141ec4549/WinUIGallery/App.xaml.cs#L26).

Jay-o-Way: PTAL

ChenYiLins:
@Jay-o-Way There are some problems with the new code. Let me explain them separately:

1. There is no point in using `GetRequiredService()`. The original `GetService<T>()` is not only functionally equivalent to `GetRequiredService`, but even throws out detailed error messages.

2. Change it to
```c#
public static MainWindow MainWindow => (Current as App)!.Host.Services.GetRequiredService<MainWindow>();
```
As a result, the DI container will re-run the search logic every time. Although  `AddSingleton` ensures that the objects are the same, it is meaningless waste.

3. 
>     // Expose the DI-created MainWindow for legacy call sites.
>     // This preserves compatibility with existing App.MainWindow usages.
>     // Consider refactoring callers to use dependency injection directly.

I don't understand what these comments explain. Does AI mistakenly think that the original method does not conform to DI dependency injection? Is `App.GetService<>()` being used in the MainWindow? But in fact, we use `public MainWindow(INavigationService navigationService)`

4. There is no difference between the new method and `null!`. Tell the compiler to ensure that it will not be null when used, and you will still throw an exception when the new method you use cannot be found. In any case, you can't use the `MainWindow` before `OnLaunched`.

Jay-o-Way:
1. Why move window creation into the activation pipeline?

WinUI 3’s activation pipeline expects the Host to be fully initialized before creating or showing the window. Moving window creation into the activation flow ensures:

- the Host is built
- all services are available
- activation handlers have run
- the window is created at the correct point in the lifecycle

This avoids inconsistent state during startup.

2. Why resolve MainWindow from DI instead of storing it statically?

In WinUI 3, the window is just another app component. Resolving it from DI:

- ensures a single, consistent instance
- avoids global static state
- makes dependencies explicit
- aligns with the recommended WinUI 3 + Generic Host pattern
- prevents accidental multiple window creations

3. Why remove App.GetService<T>()?

App.GetService<T>() acted as a global service locator. In WinUI 3, using the Host’s service provider directly is the recommended approach. Removing the static helper:

- eliminates hidden dependencies
- avoids global state
- makes the DI flow explicit
- improves maintainability and testability

```cs
    public static MainWindow MainWindow =>
        (Current as App)!.Host.Services.GetRequiredService<MainWindow>();
```
Is a "bridge solution". If you delete this bit, you will see that every `App.MainWindow.<Something>()` will fail. 
The other route would have been to rewrite all of those to use the window service.

4. Why use GetRequiredService<T>() instead of GetService<T>()?

In WinUI 3 startup, the services being resolved (MainWindow, activation handlers, core services) are mandatory. GetRequiredService<T>() ensures:

- fail‑fast behavior if a required service is missing
- no silent nulls
- predictable startup
- easier debugging

GetService<T>() is only appropriate for optional services, which these are not.

ChenYiLins:
@Jay-o-Way AI is telling you the wrong answer. I don't know what to say. o(≧口≦)o

> Why move window creation into the activation pipeline?

What does this mean?

> Why resolve MainWindow from DI instead of storing it statically?

Our MainWindow is parsed by DI, and static storage is just for easy access. (Let me give you an example. Here is a table and cola. The static property is like putting cola on the table. You can drink it whenever you are thirsty. The DI container is a supermarket, and cola can't be put on the table. Every time you are thirsty, you have to go to the supermarket. First, ask the salesman if there is any cola, and then the salesman will give it to you after checking.) Just like I wrote it differently. Can you understand it?

```c#
private static MainWindow? _mainWindow;
public static MainWindow MainWindow => _mainWindow ?? throw new InvalidOperationException("MainWindow accessed before OnLaunched.");

protected override async void OnLaunched(LaunchActivatedEventArgs args)
{
    _mainWindow = Host.Services.GetRequiredService<MainWindow>();
}
```

> GetService() is only appropriate for optional services, which these are not.

We don't use Host's GetService, but we create a method ourselves.

Jay-o-Way:
I'm sorry @ChenYiLins . Looks like we both don't understand the changes the other one has made 😔

I saw you are trying to use Dependency Injection for the MainWindow handling. I asked Copilot to explain this for me (because this really is still new and "terra incognita" for me) and after that I (or should I say "we") had some more ideas/changes to work on top of that. I simply tried to help with this DI approach. Unfortunately, that chat is no longer in my Copilot history - and I hate that too, because then I could have given the best answers.

Here's today's chat, if you want to read up on it: https://copilot.microsoft.com/conversations/join/URz5j8Fz8hbKvZf7wFLgJ

ChenYiLins:
@Jay-o-Way Ah, my apologies! I overlooked that you might not be very familiar with Dependency Injection (DI). Sorry about that! Let me give you a simple explanation.

Why did I create this PR? A long time ago, we used Template Studio to create the App project. Although its code seems very outdated to us now (as it hasn't been updated for a long time), the structure within the template was great! It built a solid foundation capable of supporting large-scale projects. However, while refactoring AutoDarkMode, in our pursuit of speed, we hardly followed the DI and MVVM patterns. That is why I am starting to modify things now—to make the project more robust, moving from "functional" to "well-crafted." Bringing these non-standard elements into alignment with best practices will be much better for the project's future development.

Now, let me explain DI. Imagine a scenario where you're buying a computer. Originally, we were picking out the motherboard components and monitors ourselves, and we had to manually select the specific models. DI is like having a concierge; we tell them "I need a monitor," and we don't need to worry about the specific model (this is the interface, like the `ICloseService` I just added).

The code we were using, `MainWindow = new MainWindow();`, is not compliant with DI. Actually, in the original template's `App.xaml.cs`, a special `GetService<T>()` was built. Its function is similar to the `GetRequiredService<T>()` you are using now; the former is something we built ourselves, while the latter is an official method belonging to the `Host`. I looked at your conversation with Copilot, and one point it misunderstood was that it assumed the `GetService<T>()` you mentioned was the official `Host` method (since `Host` also has a method called `GetService()`). Let me explain what `GetRequiredService<T>()` does: it’s like a supermarket clerk we ask, "Is there any Cola?" If there isn't, she throws an exception. The `GetService` we built ourselves not only throws an exception but also tells us what brand the Cola is.

Regarding why we use `public static Window MainWindow { get; set; } = null!;`: this is actually a bit of a "lazy" approach where we skip null checks. The AI gives the most rigorous answer, but as developers, we know that we have `Svc` as the background service, and the `App` is just a window program used to configure `Svc`. We won't have any code accessing `App.MainWindow` before `OnLaunched`, and we won't have multiple window instances, so we don't need redundant null checks.

I'll make some modifications to your code in a moment. I'm really sorry; I find explaining misunderstandings to AI quite difficult (っ °Д °;)っ, so I got a bit impatient. However, I truly appreciate your eagerness to contribute! I hope my explanation makes sense to you. There might be more PRs in the future—for instance, the MVVM is currently a mess and hasn't been properly decoupled. I also hope you can learn more as you modify the code! ヾ(≧▽≦*)o

Jay-o-Way:
Thanks. Would you like to create an issue for these* PRs, so we can track them and prepare for them?

(*open and future)

ChenYiLins:
> Thanks. Would you like to create an issue for these* PRs, so we can track them and prepare for them?
> 
> (*open and future)

I have created an issue #1222 to track these problems. I think this PR has been completed. @Jay-o-Way Do you have any more ideas?

Armin2208:
Is this ready for merging? 
Review of @Spiritreader would be also nice.

Armin2208 (approved):
will merge soon

PR #1203, opened 2026-03-11
Code quality: Use RelayCommand instead of creating ICommand and add GlobalUsings

ChenYiLins:
### Description

- Use RelayCommand instead of creating ICommand yourself. There is no performance improvement, but it will be more convenient and easy to use.
Sample code for RelayCommand:
```c#
[RelayCommand]
private void CheckUpdate()
{
    xxx;
}
```

- Add global using, but be restrained to prevent namespace conflicts.

For reviewers:

In order to facilitate the review, there is no substantial change in the code of this PR, but only changes ICommand into RelayCommand, and then cleans up the code after adding GlobalUsings.


AutoDarkModeApp/ViewModels/SettingsViewModel.cs
```diff
@@ -241,27 +243,6 @@ private async void LoadSettings()
...
-        catch (Exception)
-        {
-            throw new AddAutoStartException($"Could not validate autostart", "ValidateAutostart");
-        }
-        await Task.Delay(fakeResponsiveUIDelay);
-        SetAutostartDetailsVisibility(true);
-    }
-
-
     private async Task GetAutostartInfo(bool toggleVisibility = true)
```
Jay-o-Way, line 246: image

Spiritreader (approved):

PR #1209, opened 2026-03-18
Refactor theme collection handler

### LINKED ISSUE #1208: Allow selecting base themes as well
Allow selecting themes from the base folder `C:\Windows\Resources\Themes` as well. Currently, the program does include default themes in its listing, but any additional themes added to that folder will not be recognized. The only workaround is to build a new custom theme using the new theme, then save it to `C:\Users\[User]\AppData\Local\Microsoft\Windows\Themes` for it to be selectable in the program.

Jay-o-Way:
fixes #1208

Refactored the handler to search for .theme files, also in the Windows Themes folder.
Note that the `DisplayName` is not a friendly string inside the built-in Windows themes, so I used a look-up table to connect the strings from resources.
Moved the **InjectWindowsThemes** code into the larger code block.

✔️ Somebody check if this is correct:

```cs
 if (file.Contains(Helper.PathUnmanagedDarkTheme) ||
    file.Contains(Helper.PathUnmanagedLightTheme) ||
    file.Contains(Helper.PathManagedTheme)) continue;
```

Because a different part ~uses~ used "NameUnmanagedLightTheme"

```cs
 .Where(f => !f.Contains(Helper.PathUnmanagedDarkTheme) && !f.Contains(Helper.NameUnmanagedLightTheme) && !f.Contains(Helper.PathManagedTheme)).ToList();
```

~❓ Question: should we (alphabetically) sort the final list?~ done

> [!NOTE]
> I don't use Windows 10 anymore, so would be great if somebody could test that! 

### Screenshot? Sure!

Theme switching

### DIFF
--- a/AutoDarkModeApp/Strings/en-us/Resources.resw
+++ b/AutoDarkModeApp/Strings/en-us/Resources.resw
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!-- 
+    Microsoft ResX Schema 
+    
     Version 2.0
-
-    The primary goals of this format is to allow a simple XML format
-    that is mostly human readable. The generation and parsing of the
-    various data types are done through the TypeConverter classes
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
     associated with the data types.
-
+    
     Example:
-
+    
     ... ado.net/XML headers & schema ...
     <resheader name="resmimetype">text/microsoft-resx</resheader>
     <resheader name="version">2.0</resheader>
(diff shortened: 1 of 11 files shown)


ChenYiLins:
Nice PR! Respond quickly to Issues!

> ✔️ Somebody check if this is correct:
> 
> ```cs
>  if (file.Contains(Helper.PathUnmanagedDarkTheme) ||
>     file.Contains(Helper.PathUnmanagedLightTheme) ||
>     file.Contains(Helper.PathManagedTheme)) continue;
> ```
> 
> Because a different part ~uses~ used "NameUnmanagedLightTheme"

No, this part of the code is incorrect, because all three variables correspond to complete paths, not file names. You can use NameUnmanagedLightTheme instead, and add a NamePathManagedTheme for PathManagedTheme, which can avoid hard coding.

> ❓ Question: should we (alphabetically) sort the final list?

I think it's a good idea. In the case of multi-theme files, you can quickly find the theme files you want.

> I don't use Windows 10 anymore, so would be great if somebody could test that!

As far as the current changes are concerned, it should not affect the compatibility of Win10, because as far as I know, Win11 has little change in theme compared with Win10.

(In the next few days, I can't use the computer for personal reasons, so Jay, I can't help you change this PR, I can only look at it. （；´д｀）ゞ)

Jay-o-Way:
EnumerateFiles always returns full paths, not just filenames. So the safety thing would be to compare paths against paths, right? Also be using `Equals(...)`

Jay-o-Way:
image
Done

Spiritreader (approved):

Armin2208 (approved):

Issue #1190, opened 2026-01-28
[Ambient Light] Skip auto switch one should be hidden from ComboBox


Armin2208:
While Ambient Light Sensor Switch is enabled, it is not possible to skip the auto switch ones. This makes sense, because we don't have any time to rely on.

BUT the option is still viewable on the Time Page User Interface. The combobox should not reveal this option while the AmbientLightGovernor is enabled.

ChenYiLins:
It is not difficult to get this function. 

> This makes sense, because we don't have any time to rely on.

However, I found that if we follow this reasoning, we should not enable this option in Windows night mode, but in fact, this option is available, that is, we simply make the next switch not take effect., just like this:

<img width="865" height="175" alt="Image" src="https://github.com/user-attachments/assets/562b3d61-03c6-411d-b171-dc3086866ed9" />

In the same way, this should be feasible in ambient light mode. So we shouldn't hide it?

Armin2208:
This option makes sense on the Night Light Mode, because it follows a time schedule, which switches the theme two times per day. If you skip your auto switch once, you skip one time dark theme.  That makes sense to me. Its the same just like you would control the time via Auto Dark Mode, just the timer is external.

The ambient light sensor maybe switches multiple times per day. So if a user don't want a switch, because he is working on the train with multiple dark/light switches, the "skip auto switch once" option doesn't make much sense.

Maybe it's just me, but I think it doesn't make a lot of sense in general for the Ambient Light Sensor.
Maybe you could argue: if someone is using his notebook just at home, he only has a switch if he turns off the light or the sun goes down. So the auto skip once would help for that.
But still: I am kinda thinking the Ambient Light Feature is not the right place for a skip auto switch once.

And I am not sure anymore, but my wording suggests that the skip auto once isn't even working while Ambient Light is activated:
"While Ambient Light Sensor Switch is enabled, it is not possible to skip the auto switch ones. "

ChenYiLins:
> The ambient light sensor maybe switches multiple times per day. So if a user don't want a switch, because he is working on the train with multiple dark/light switches, the "skip auto switch once" option doesn't make much sense.

This reason successfully convinced me. **Skip auto switch once** is feasible, but this option is not reasonable.

Let's merge that PR. φ(*￣0￣)

Armin2208:
Great 😂🫶🏻

PR #1223, opened 2026-05-01
Fix: Remove skip option from combobox in ambient light mode

ChenYiLins:
### Description

As the title indicates.

But I personally have some questions about this Fix. Let's track these questions in #1190. If we deny this Fix in the discussion, we will close this PR.

### Screenshots

image


Armin2208 (approved):
thank you

PR #1224, opened 2026-05-02
Fix: Disable duplicate shortcut keys

### LINKED ISSUE #1146: [11.1] Don't allow multiple Hotkeys having the same Hotkey
Image

ChenYiLins:
### Description

Before, we allowed multiple shortcuts to use the same shortcut key. This PR improves this problem. Now when you try to save the same shortcut key, an error message will pop up and you can't save it.

Sloved #1146.

### Screenshots

image

### DIFF
--- a/AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml
+++ b/AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml
@@ -10,6 +10,13 @@
 
     <StackPanel Orientation="Vertical" Spacing="16">
 
+        <InfoBar
+            x:Name="ErrorInfoBar"
+            IsClosable="False"
+            IsOpen="{x:Bind IsErrorVisible, Mode=OneWay}"
+            Message="{x:Bind ErrorMessage, Mode=OneWay}"
+            Severity="Error" />
+
         <TextBlock Text="{helpers:ResourceString Name=SelectTheKeyCombination}" TextWrapping="Wrap" />
 
         <ItemsControl
--- a/AutoDarkModeApp/Views/HotkeysPage.xaml.cs
+++ b/AutoDarkModeApp/Views/HotkeysPage.xaml.cs
@@ -42,18 +42,30 @@ public sealed partial class HotkeysPage : Page
             Content = dialogContent,
         };
 
+        shortcutDialog.Closing += (dialog, args) =>
+        {
+            if (args.Result == ContentDialogResult.Primary)
+            {
+                var (isDuplicate, conflictingName) = ViewModel.IsDuplicateHotkey(hotkeyData.Tag, dialogContent.CapturedHotkeys);
+                if (isDuplicate)
+                {
+                    dialogContent.ShowError(string.Format("HotkeyDuplicateErrorMessage".GetLocalized(), conflictingName));
+                    args.Cancel = true;
+                }
+            }
+        };
+
         var result = await shortcutDialog.ShowAsync();
         if (result == ContentDialogResult.Secondary)
         {
             dialogContent.HotkeyCombination?.Clear();
             dialogContent.CapturedHotkeys = null;
         }
-        else if (result != ContentDialogResult.Primary)
-        {
-            return;
-        }
 
-        ViewModel.UpdateHotkeyValue(hotkeyData.Tag, dialogContent.CapturedHotkeys);
+        if (result == ContentDialogResult.Primary || result == ContentDialogResult.Secondary)
+        {
+            ViewModel.UpdateHotkeyValue(hotkeyData.Tag, dialogContent.CapturedHotkeys);
+        }
     }
 
     private async void SaveSettingsButton_Click(object sender, RoutedEventArgs e)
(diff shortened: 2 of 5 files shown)


AutoDarkModeApp/Views/HotkeysPage.xaml.cs
```diff
@@ -42,18 +42,30 @@ private async void EditHotkeysButton_Click(object sender, RoutedEventArgs e)
             Content = dialogContent,
         };
 
+        shortcutDialog.Closing += (dialog, args) =>
+        {
+            if (args.Result == ContentDialogResult.Primary)
+            {
+                var (isDuplicate, conflictingName) = ViewModel.IsDuplicateHotkey(hotkeyData.Tag, dialogContent.CapturedHotkeys);
+                if (isDuplicate)
```
Jay-o-Way: Can we move this to the "changed" `OnKeyboardEvent` event, for immediate display? Similar to the check for a "valid" hotkey.

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml
```diff
@@ -10,6 +10,13 @@
 
     <StackPanel Orientation="Vertical" Spacing="16">
 
+        <InfoBar
+            x:Name="ErrorInfoBar"
+            IsClosable="False"
+            IsOpen="{x:Bind IsErrorVisible, Mode=OneWay}"
+            Message="{x:Bind ErrorMessage, Mode=OneWay}"
```
Jay-o-Way, line 17: Do we re-use this InfoBar for different messages? If not, why ~bind to~ have a variable?

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
```diff
@@ -42,18 +42,30 @@ private async void EditHotkeysButton_Click(object sender, RoutedEventArgs e)
...
+
         var result = await shortcutDialog.ShowAsync();
         if (result == ContentDialogResult.Secondary)
         {
             dialogContent.HotkeyCombination?.Clear();
             dialogContent.CapturedHotkeys = null;
         }
-        else if (result != ContentDialogResult.Primary)
+
+        if (result == ContentDialogResult.Primary || result == ContentDialogResult.Secondary)
```
Jay-o-Way, line 68: Unless I'm mistaken, Secondary == Cancel == blank hotkey

ChenYiLins:
## Code quality: Show conflict results immediately

## Fix: Disable invalid shortcut keys

### Description

- Now the key conflict will be displayed immediately.
- Shift and Ctrl will be displayed immediately (previously, they were not displayed when only the Shift or Ctrl key was pressed)
- Saving invalid shortcut keys (only Shift or Ctrl) will not be allowed.

### Screenshots

image

image

PixPin_2026-05-03_17-02-45

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
ChenYiLins: Done. (By the way, I used delegation to realize this function, but it also led to the problem of higher code coupling, but I really couldn't think of a better solution, because the key information was saved in the ViewModel of the page, which had nothing to do with ShortcutDialogContentControl, but the InfoBar of error information was in ShortcutDialogContentControl)

AutoDarkModeApp/UserControls/ShortcutDialogContentControl.xaml
ChenYiLins: Yes, we let the repeated keys be displayed.

```c#
ShowError(string.Format(ResourceExtensions.GetLocalized("HotkeyDuplicateErrorMessage"), conflictingName));
```

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
ChenYiLins: No, the SecondaryButton is Reset, so we need to refresh the edit box outside after clicking the reset button. So we need this line of code.

Armin2208:
>     * Shift and Ctrl will be displayed immediately (previously, they were not displayed when only the Shift or Ctrl key was pressed)
>     * Saving invalid shortcut keys (only Shift or Ctrl) will not be allowed.

Really good and important additions, thanks.

AutoDarkModeApp/Views/HotkeysPage.xaml.cs
Armin2208: I actually don't like the OnKeyboardEvent for the new addition of showing that only modifier keys aren't allowed. Now as soon the user presses one button (for example ctrl) he sees a red message. I think its irritating, even though it's only for one second.
So at least for the Key Modifier warning, it should only appear if the shortcut dialog wants to close.

ChenYiLins: > I actually don't like the OnKeyboardEvent for the new addition of showing that only modifier keys aren't allowed. Now as soon the user presses one button (for example ctrl) he sees a red message. I think its irritating, even though it's only for one second. So at least for the Key Modifier warning, it should only appear if the shortcut dialog wants to close.

Done!

Armin2208 (approved):

Spiritreader (approved):
Thanks Chen! Nice addition

PR #1238, opened 2026-07-18
Feature: Support energy saving mode switch theme

ChenYiLins:
### Description

- Done one of #15

When the computer's power-saving mode is turned on, the theme switches to dark mode; when power-saving mode is turned off, it switches back to light mode.

### Supplement

The reason for not using the WinRT API is that, following the [Windows 11 24H2 update](https://learn.microsoft.com/en-us/windows-hardware/design/component-guidelines/energy-saver), the new power-saving mode features have been expanded, while `PowerManager.EnergySaverStatus` still uses the old battery power-saving mode. For more details, please refer to this [link](https://learn.microsoft.com/en-us/answers/questions/5579583/powermanager-energysaverstatus-not-working-properl). Testing has shown that when the device is plugged in, this value is always set to “Disabled.” The Win32 API behaves the same way. The registry solution, however, offers greater versatility, which is the compatibility that ADM has always sought.

### Screenshots

image

PixPin_2026-07-18_16-10-14

### DIFF
--- a/AutoDarkModeApp/Strings/en-us/Resources.resw
+++ b/AutoDarkModeApp/Strings/en-us/Resources.resw
@@ -1125,4 +1125,7 @@ Please note: This may introduce slight to moderate lag on some systems during th
   <data name="Contributors" xml:space="preserve">
     <value>Contributors</value>
   </data>
+  <data name="FollowEnergySavingMode" xml:space="preserve">
+    <value>Follow energy saving mode</value>
+  </data>
 </root>
--- a/AutoDarkModeApp/Strings/en-us/Resources.resw
+++ b/AutoDarkModeApp/Strings/en-us/Resources.resw
@@ -1126,6 +1126,9 @@ Please note: This may introduce slight to moderate lag on some systems during th
     <value>Contributors</value>
   </data>
   <data name="FollowEnergySavingMode" xml:space="preserve">
-    <value>Follow energy saving mode</value>
+    <value>Follow Energy Saver mode</value>
+  </data>
+    <data name="OpenWindowsPowerSleep" xml:space="preserve">
+    <value>Go to Power sleep settings</value>
   </data>
 </root>
--- a/AutoDarkModeApp/ViewModels/TimeViewModel.cs
+++ b/AutoDarkModeApp/ViewModels/TimeViewModel.cs
@@ -20,6 +20,7 @@ public partial class TimeViewModel : ObservableRecipient
         CoordinateTimes,
         WindowsNightLight,
         AmbientLight,
+        PowerLight
     }
 
     [ObservableProperty]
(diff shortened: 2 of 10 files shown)


AutoDarkModeApp/Strings/en-us/Resources.resw
```diff
@@ -1125,4 +1125,7 @@ Please note: This may introduce slight to moderate lag on some systems during th
   <data name="Contributors" xml:space="preserve">
     <value>Contributors</value>
   </data>
+  <data name="FollowEnergySavingMode" xml:space="preserve">
+    <value>Follow energy saving mode</value>
```
Jay-o-Way: Since it's a named Windows feature `Energy Saver` should have capital letters

AutoDarkModeApp/ViewModels/TimeViewModel.cs
```diff
@@ -20,6 +20,7 @@ public enum SwitchTriggerMode
         CoordinateTimes,
         WindowsNightLight,
         AmbientLight,
+        PowerLight
```
Jay-o-Way: "light" has nothing to do with this feature :)

Jay-o-Way (commented):
Seems to work fine on my computer. But I don't have a laptop, so I can't test battery related things.
One thing I would like to add: a link to Windows Energy Saver `ms-settings:powersleep` settings at the bottom of the Radio Buttons.

Jay-o-Way:
What if user has Windows 10 and no battery? Should the setting be disabled? 🤔

ChenYiLins:
> Seems to work fine on my computer. But I don't have a laptop, so I can't test battery related things. One thing I would like to add: a link to Windows Energy Saver `ms-settings:powersleep` settings at the bottom of the Radio Buttons.

Great suggestion! It's already done!

Armin2208:
Thanks for working on this.
I actually would suggest that this isn't an own Governor, instead a module. Because we already have for example dark mode while on battery as module.

Usually Energy Saver Mode will be only enabled by user on low battery. So as long as they are on a high battery percentage, they probably still want to use time based dark mode.

Jay-o-Way:
> Usually Energy Saver Mode will be only enabled by user on low battery. So as long as they are on a high battery percentage, they probably still want to use time based dark mode.

Assumptions, Armin. Be careful with them. Energy saver is made available on computers that don't have a battery.

ChenYiLins:
> Thanks for working on this. I actually would suggest that this isn't an own Governor, instead a module. Because we already have for example dark mode while on battery as module.
> 
> Usually Energy Saver Mode will be only enabled by user on low battery. So as long as they are on a high battery percentage, they probably still want to use time based dark mode.

Yep, I saw it. We already have a module in place, and what I’m considering now is having the system automatically switch to dark mode when a user manually enables power-saving mode. This is indeed a feature that users can enable themselves, but it’s not a very common idea. So let’s put this PR on hold for now; I’ll try to implement the screen brightness switch instead.

Spiritreader:
@Armin2208 
> Thanks for working on this. I actually would suggest that this isn't an own Governor, instead a module. Because we already have for example dark mode while on battery as module.

Yeah I think this makes sense. A module would be able to both work when routine-based switching (via governor) is enabled, and also be able to serve as stand-alone.

@Jay-o-Way 
> Assumptions, Armin. Be careful with them. Energy saver is made available on computers that don't have a battery.

What Armin said is neither in conflict with that nor does it invalidate the design decision

Armin2208:
resolved in #1262

Issue #1200, opened 2026-03-08
TO-DO list of converting translation workflow


Jay-o-Way:
### Situation

We encounter a lot of merge problems when Weblate creates a PR to give us translations, because we have all the resource files in the repository, and all of them can be modified during development. Thanks to Visual Studio (editor) applying changes to all those files.

### Goal

A clean workflow, where we automate what we can, to prevent merge conflicts, and a smooth and safe way to include translations in the build. In short: remove all translations from the repo, let Weblate handle translations, let CI Build pull them when needed.

### Workflow

1. GitHub Repository Setup
   1.1 Keep only English resources in the repo
   1.2 Add .gitignore rule to prevent translated files from being committed
   1.3. Add a Git pre-commit hook to block accidental edits
   1.4. Document the rule: “Developers edit only en-US”
2. Weblate Setup
   2.1. Create Weblate components **(see comment below)**
   2.2. Set Weblate to track the active development branch
   2.3. Enable authoritative translation mode
   2.4. Disable Weblate PRs
   2.5. Set target languages (Weblate → Project → Languages)
3. CI Pipeline Setup (GitHub Actions or Azure DevOps)
   3.1. Add a step to download translations from Weblate
   3.2. Copy translations into the build folder
   3.3. Build the app with translations included
   3.4. Package and publish
4. Release Workflow
   4.1. When releasing a new version
   4.2. Switch Weblate to the new development branch
5. Optional Enhancements
   - Cache translations in CI
     Avoid hitting Weblate API too often
   - Add a “translation status” badge
   - Add automated checks
     Fail build if English .resw is malformed.
   - Add a local script to pull translations from Weblate
     To be able to test languages locally

💬 Full conversation: https://copilot.microsoft.com/conversations/join/y4WVraaV3yAJ3w59LsULq
📝 Full list: [Translation Workflow.pdf](https://github.com/user-attachments/files/25848644/Translation.Workflow.pdf)
🔗 Link: https://hosted.weblate.org/projects/auto-dark-mode/

Jay-o-Way:
### Note

Since we have different projects with different directory structures for Strings, we need to remember that when setting up Weblate: **One Weblate component per project**. The good news is: **you do NOT need to restructure your SVC project**, and the CI pipeline can still inject translations correctly — even if the folder structure differs between projects.

> • 	You do not need to change the SVC project structure
• 	You create one Weblate component per project
[...]

PR #1242, opened 2026-07-25
Persian Translation

notadev-lol-91:
Added some new Persian Translations (And converted all line endings to CR+LF)
(By the way I'm not really a developer and would prefer to not become a regular contributor to this project)


Jay-o-Way:
Hi there! Thank for the effort. This would make a lot of people happy.

Unfortunately, this is not the place to add these translations. We use Weblate to handle translations, and we have the idea to change the whole translation procedure some time in the future. See #1200 for reference.

Would you like to add your translations to https://hosted.weblate.org/projects/auto-dark-mode/app/fa/? that would be lovely! 💙

notadev-lol-91:
Sorry for the misunderstanding because it said so in your `Readme`! Oh well. I'll close the PR But keep the fork. _(Maybe add it to weblate if it comes up from maintenance)_

Jay-o-Way:
@Spiritreader orrr is this an exception?

PR #1245, opened 2026-07-26
Refactor TimePage to AutoSwitchPage

### LINKED ISSUE #1244: Task: rename and refactor TimePage
As we know, we have options that do not depend on the time. Suggest to rename the page (and code-behind) to `AutoSwitchPage`.

I also want to change the UI to use **SettingsExpander** controls with the **ToggleSwitch** as the main settings (i.e. the **Header**) and the rest below that. _Visibility_ must be be set on each **SettingsCard**.

Also, the **ViewModel** (file) is absolutely huge and a big mess. It needs to be split in a few "partial" files and be properly sorted out. It's a wonder it works as well as it does now.

---

Optional: Some of the RadioButton controls are optional - they depend on hardware. Best practice is to HIDE impossible settings, but the **RadioButtons** (container) control preserves space if a RadioButton has `Visibility=Collapsed`. Solution is to use `<RadioButtons ItemsSource="{x:Bind ViewModel.SwitchModes}"...>`, where items may or may not be _added_, but that means a large refactor.

Jay-o-Way:
HUGE refactor of the whole page and code, including the "postpone" part to "pause".

- Renamed the Page title and code-behind
- Using **SettingsExpander** and **SettingsCard** controls for clean UI
- Moved the SettingsCard sections in a new order: input first, output last
- Merged two identical SettingsCard controls
  see `Visibility="{x:Bind ViewModel.LocationSettingsCardVisibility, Mode=OneWay}"`
- Split the **ViewModel** (Core, Triggers, TimeLocation, AmbientLight, Pause...)
- Completely renamed and rebuilt the "postpone" section to "pause"
  - Now using one single **ComboBox** for all the options: Off, Once, Timed
  - More code to handle **ComboBoxItem** controls, and handle all the logic
  - ...this includes hiding the "Once" option when trigger = AmbientLight
- Renamed some methods
- added `Pause.Once` mode in the back (to prevent ComboBox being set to the last item)

### To do

Merge conflicts. I **hate** merge conflicts😠

refactor TimePage


AutoDarkModeApp/ViewModels/AutoSwitchViewModel.Pause.cs
```diff
@@ -0,0 +1,229 @@
...
+                        }
+                        return options.Length + 1; // If greater than all options, return the last index (8)
+                    }
+
+                    int desiredIndex = result.Mode switch
+                    {
+                        PauseMode.Off => 0,
+                        PauseMode.Once => 1,
+                        PauseMode.Timed => MapMinutesToIndex(result.Minutes),
+                        _ => 0
```
Jay-o-Way, line 139: Something isn't right here. Probably a ~racing~ model mismatch issue between front and back.
Repro: select "Once" in the ComboBox. Result: last **Item** is selected.

AutoDarkModeApp/ViewModels/AutoSwitchViewModel.Core.cs
```diff
@@ -0,0 +1,292 @@
...
+        {
+            SelectedTriggerMode = SwitchTriggerMode.CoordinateTimes;
+        }
+
+        OffsetTimesMinimum = -720;
+        TimePickerVisibility = Visibility.Visible;
+        OffsetTimeSettingsCardVisibility = Visibility.Visible;
+    }
+
+    private static async void SafeApplyTheme()
```
Jay-o-Way: This is horrible code. It's not safe, it's not apply, it's not theme. Need to rename: RequestThemeAsync. Need to use Task and check result. Also: FIFTEEN seconds?! Who?

AutoDarkModeApp/ViewModels/AutoSwitchViewModel.Commands.cs
```diff
@@ -0,0 +1,56 @@
+using System.Globalization;
+
+namespace AutoDarkModeApp.ViewModels;
+
+// Relay Commands
+public partial class AutoSwitchViewModel : ObservableRecipient
+{
+    [RelayCommand]
+    private void AutoConfigure()
```
Jay-o-Way: Too vague. Rename to AutoConfigureLightSensor

AutoDarkModeApp/ViewModels/AutoSwitchViewModel.Pause.cs
```diff
@@ -0,0 +1,224 @@
...
+                    result.InfoText = "Statuscode: Timeout";
+                    _isInitializing = false;
+                    return;
+                }
+                // Disabled
+                else if (reply.StatusCode == StatusCode.Disabled)
+                {
+                    result.Mode = PauseMode.Off;
+                    result.Minutes = null;
+                    //PauseInfoText = "Msg_AutoSwitchDisabled".GetLocalized();
```
Jay-o-Way: 🧹

Jay-o-Way (commented):
Not perfect yet

AutoDarkModeApp/ViewModels/AutoSwitchViewModel.Pause.cs
ChenYiLins: The front-end logic is incorrect. Skipping once should not be a special mode; it also falls under `PauseMode.Timed`, but `CurrentPauseMinutes = 0;` should be used instead.

ChenYiLins:
I don’t object to splitting up the ViewModel, since the previous TimeViewModel was indeed too verbose. However, isn’t the current breakdown a bit excessive? We could try merging some business logic into the Services. Also, the current breakdown is a bit confusing—for example, the Commands don’t include all commands. There should be a consistent standard for splitting, rather than doing it arbitrarily, as this will lead to inconsistencies in the coding style of ViewModels across other pages.

Jay-o-Way:
Thanks for the feedback! I agree that splitting ViewModels should follow a consistent standard — and that’s exactly why this PR uses domain‑based boundaries rather than arbitrary file size or service grouping.

The four ViewModels correspond directly to the four functional domains in the XAML: switching logic, pause logic, location/sun times, and ambient light. Each domain has its own UI state, its own backend interactions, and its own reentrancy concerns. Keeping these concerns isolated is what finally makes the page predictable and testable.

Moving business logic into Services would blur the MVVM separation: Services are stateless and backend‑focused, while ViewModels manage UI state and reflection. If Services start handling UI‑state logic (e.g., PauseInfoText, lux descriptions, slider mapping, initialization guards), we end up with inconsistent patterns across pages and harder‑to‑test code.

So the goal here is not “more files”, but “clear domain boundaries”. This structure scales well to other pages because it provides a repeatable pattern: one ViewModel per functional domain, with a simple root ViewModel orchestrating them.

If you’d like, I can document this pattern so future pages follow the same structure.

AutoDarkModeApp/ViewModels/AutoSwitchViewModel.Pause.cs
Jay-o-Way: You are correct: the frontend has a concept "Once" that doesn't exist in the backend, so LoadPauseTimer() reads what you mean.

~I can update the backend to have `PauseMode.Once`, this seems (relatively) easy~

Armin2208:
As I already state in the corresponding issue, I really like this design.

Here some notes and thoughts:
- Auto switch is a good name for this page.
- Is it possible to move the geolocation input area above the time view? Because as a user, if you set up Auto Dark Mode, you press on "geographic coordinates" and the next step would be to put in the Latitude and Longitude. Its irritating having these controls more to the bottom. 
- Also, additionally for the geolocation input area: I don't like that the button "Find geo coordinates for your location" is seperated from the textbox. I know, Microsoft also sometimes does stuff like this, but I think they should be paired together, because they are also logically bounded.
- We should remove the 15 Minutes pause option, because it is a really short time. We should start at 30 minutes. It should bes till a combo box though, because this is user friendly. I would also rename "once" to: "Until Sunrise" / "until sunset" just like in tray.
- Auto-set thresholds is a nice idea, but I like more the word calibrate. Auto calibrate thresholds?

Also, and maybe most important, the titles are doubled. Above switch behavior the title says switch behavior.

Jay-o-Way:
Hey @Armin2208, good to see from you again! Thanks for the comments, too!

> * Auto switch is a good name for this page.

Yay 😀

> * Is it possible to move the geolocation input area above the time view? Because as a user, if you set up Auto Dark Mode, you press on "geographic coordinates" and the next step would be to put in the Latitude and Longitude. Its irritating having these controls more to the bottom.

Of course it is possible, and from a "chronological" point of view that makes sense. with that in mind, the **TimePicker** controls should always be the last in line. ✔️

> * Also, additionally for the geolocation input area: I don't like that the button "Find geo coordinates for your location" is seperated from the textbox. I know, Microsoft also sometimes does stuff like this, but I think they should be paired together, because they are also logically bounded.

Easy to do. And you're actually right to note this, because in a few places in Windows Settings a hyperlink is actually placed right in the same SettingsCard, exactly like you suggest! ✔️

> * We should remove the 15 Minutes pause option, because it is a really short time. We should start at 30 minutes. It should bes till a combo box though, because this is user friendly.

Sure, easily done. ✔️

> I would also rename "once" to: "Until Sunrise" / "until sunset" just like in tray.

One string, so easy to change. **However**, this could sound like a pause UNTIL - but not including - the next sunrise/sunset time. While in fact it SKIPS the first next event, and turns back on at the next event after that. Therefore I think "Skip once" would be better.

> * Auto-set thresholds is a nice idea, but I like more the word calibrate. Auto calibrate thresholds?

I'm convinced the word "calibrate" is not appropriate. It implies something is compared to a standard, and that's not the action.

> Also, and maybe most important, the titles are doubled. Above switch behavior the title says switch behavior.

Done. ✔️

AutoDarkModeSvc/Core/PostponeManager.cs
```diff
@@ -197,6 +197,26 @@ public bool ToggleSkipNextSwitch()
...
+            Remove(existing.Reason);
+            return false;
+        }
+
+        PostponeItem item = new(Helper.PostponeItemPauseOnce, isUserClearable: true);
+
+        var (nextSwitchAdjusted, skipType) = GetSkipNextSwitchExpiryTime();
+        item.SkipType = skipType;
+
+        Add(item);
```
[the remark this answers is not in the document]
Spiritreader: Why?

toggle skip once and toggle skip next are semantically the same, they should differ depending on the governor yes, but I think it's wrong to expose two options. It's also bad API design to have two options (--toggle-skip-next) and --toggle-skip-once where the frontend app needs knowledge of which mode is active to effectively use it.

AutoDarkModeSvc/Communication/MessageParser.cs
```diff
@@ -324,11 +324,24 @@ public static void Parse(List<string> msg, Action<string> SendResponse, Service
                 #region ToggleSkipNext
                 case Command.ToggleSkipNext:
                     Logger.Info("signal received: toggle skip next switch");
-                    bool enabled = state.PostponeManager.ToggleSkipNextSwitch();
+                    bool enabledNext = state.PostponeManager.ToggleSkipNextSwitch();
```
Spiritreader, line 327: One does not simply introduce a new API endpoint without justifying.

Spiritreader (changes requested):

AutoDarkModeSvc/Communication/MessageParser.cs
ChenYiLins: Sam is actually right. I don't know if the original intention was to introduce a new logic, but the current logic appears to be in a permanent pause state, and this is also one of the reasons for the front-end display error.

Jay-o-Way:
Well, this was one hell of a lesson 🤓

PR #1248, opened 2026-08-14
Fix: NavigationView Header gets cut off

ChenYiLins:
### Description

This PR fixes a bug in WinUI itself where the navigation bar header was cut off; compared to the previous PR, this should be a more effective solution.

### Screenshots

4f78b8cd58936a9ad08b53536e482282

### Supplement

I think it’s the right approach not to restrict the navigation bar. Even though we’ve limited the window size, who knows if someone might use some weird resolution or scaling level? (っ °Д °;)っ

### DIFF
--- a/AutoDarkModeApp/Services/NavigationService.cs
+++ b/AutoDarkModeApp/Services/NavigationService.cs
@@ -36,6 +36,7 @@ public class NavigationService : INavigationService
     {
         _navigationView = navigationView;
         _navigationView.SelectionChanged += OnSelectionChanged;
+        _navigationView.DisplayModeChanged += DislplayModeChanged;
 
         _navigationView.PointerPressed += NavigationView_PointerPressed;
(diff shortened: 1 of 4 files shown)


Spiritreader (approved):

Spiritreader:
thanks @ChenYiLins!

AutoDarkModeApp/Services/NavigationService.cs
```diff
@@ -91,6 +92,24 @@ private void OnSelectionChanged(NavigationView sender, NavigationViewSelectionCh
         }
     }
 
+    private void DislplayModeChanged(NavigationView sender, NavigationViewDisplayModeChangedEventArgs args)
```
Jay-o-Way, line 95: There is a typo in the name

Jay-o-Way (commented):
If the Nav.Header is the control with the issue, why aren't we fixing the actual header? Why not use a VisualState manager?

Issue #1249, opened 2026-08-14
UI language issue after latest update


Ctrl-Alb:
### Description

Hello,
Thanks for the awesome work you’re doing with ADM!

It’s usually rock-solid, but after updating to the latest version, I noticed that most of the UI is now displayed in Chinese, even though I have “Italian” selected as the app language.
Some parts are still in English, and only a very small portion is actually translated into Italian.

### Expected Behavior

UI should be displayed in Italian

### Log Data

[service.log](https://github.com/user-attachments/files/31069687/service.log)

[config.yaml](https://github.com/user-attachments/files/31069692/config.yaml)

### Commit Hash, Version and Windows Build

- Commit: `c15b28e9`
- Service/App: `11.1.0.23`
- Updater: `4.0.3`
- Shell: `1.3.3.0`
- .Net: `9.0.11`
- Windows: `26200.9168`
- Arch: `X64`

### Screenshots / Videos

<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/869235a2-e990-4b4c-8811-87cf74f2bacb" />

Armin2208:
I didn't test the language switch behavior in my QA session for 11.1. This seems to be a mistake by myself.

Because I can see following behavior:
- After switching the language and clicking on "restart", the AutoDarkModeApp doesn't restart itself. The user needs to re-open it from the tray area. It should re-open by itself.
- After re-opening the AutoDarkModeApp after a language switch, the Window size is maximized. It seems like something while saving doesn't work.
- My UI language gets always reverted to english. I actually can't change the language on my desktop PC.
- The service reliable switches the language.

Ctrl-Alb:
No worries! As I said, ADM has usually been bug-free in my experience.

Please let me know if I can provide any additional information or test anything on my end

Jay-o-Way:
I can reproduce as in the original comment. Can't reproduce the additional problems Armin mentions. It's strange that this seems to be an issue (mostly or only) with Italian and most other languages work fine...

Spiritreader:
so basically, the italian translation was somehow overwritten by the chinese one. I've fixed that and also rebased with the current weblate data

ChenYiLins:
@Spiritreader It looks like you've already solved this problem; I just figured out where the problem was, too. (●'◡'●)

Armin2208:
I can confirm, italian is now italian in the latest beta release.
[...]

PR #1253, opened 2026-08-18
Fix: Can't change app language and unstable restarts

ChenYiLins:
### Description

This PR resolves issue #1249 and the window positioning issue that Armin previously brought to my attention.

- Uses the Windows App SDK’s `ApplicationData` API to save settings
- Uses the Windows App SDK’s `Restart` API to restart the app

I have now upgraded the Windows App SDK to version 2.0 so that we can use `ApplicationData` in unpackaged mode; this eliminates the need to use JSON to save app settings. I've added an extra method that automatically upgrades users' old profiles to the new version. I think this method can remain in place until the next major version update. I’ve also implemented `Restart` for direct app restarts.

Previously, there were three issues: the language couldn’t be changed, the window position was unstable, and the app failed to restart. However, these issues weren’t caused by the same root cause; the first two were due to problems with the code related to file writing. I was unable to pinpoint the exact cause of the last issue and can only speculate that a problem occurred with an asynchronous task. This is because, in managed mode debugging, I couldn’t locate exceptions originating within WinUI, and in native debugging, the output was flooded with WinUI-specific exceptions, leaving me with no clear starting point.

I tried to improve the asynchronous task related to language settings. The original `Task<string> GetDefaultLanguageAsync()` caused a deadlock on the UI thread, though I’m not entirely sure why. The reason the app couldn’t restart reliably was also due to this UI thread deadlock, which prevented WinUI from closing the window in time—even though a new process had already been launched. `CheckAppMutex` would then terminate the new process, preventing the app from truly restarting. In contrast, `Environment.Exit(0)` directly kills the frozen WinUI thread, closing the app without any lag, so there’s no issue with the mutex. The same principle applies to `AppInstance.Restart`, but both methods prevent the window from triggering the **Close** event, so we need to manually enable the **Close service**.


Jay-o-Way:
I thought @Spiritreader already fixed the language issue. 

And as i already said to Sam: when I wrote these bits of code, it was working fine. So what changed **after** that time? Maybe **there** lies the issue.

ChenYiLins:
> And as i already said to Sam: when I wrote these bits of code, it was working fine. So what changed **after** that time? Maybe **there** lies the issue.

No, the second and third issues I mentioned actually occurred quite a while ago (there were early signs of these problems as far back as last year, though at the time I wasn’t sure whether they were caused by WinUI itself or by our code); we’ve simply been using some workarounds to prevent these issues from affecting users.

For example, here’s an early instance of forcibly terminating the application:
```c#
        try
        {
            Process.GetCurrentProcess().Kill();
        }
        catch (Exception ex)
        {
            Debug.WriteLine(ex.Message);
        }
```
For a while, I thought we had resolved these issues, but with some updates, they resurfaced.

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
```diff
@@ -444,12 +429,6 @@ partial void OnSelectedLanguageChanged(string value)
         });
     }
 
-    private async Task SaveLanguageSettingsAsync(string languageCode, bool languageChanged)
-    {
-        await _localSettingsService.SaveSettingAsync("SelectedLanguageCode", languageCode);
-        await _localSettingsService.SaveSettingAsync("LanguageChanged", languageChanged); // used for ActivationService > jumplist
```
Jay-o-Way, line 450: Can you confirm the jumplist (etc.) is in the correct language after restart?

Jay-o-Way (commented):
> this eliminates the need to use JSON to save app settings. 

- Does that count as a breaking change? What's the new version number going to be?
- Will a user still be able to change data directly?
- Remember this needs to be documented.

AutoDarkModeApp/ViewModels/SettingsViewModel.cs
Armin2208: Yes, that actually works.

ChenYiLins:
> * Does that count as a breaking change? What's the new version number going to be?

This isn't really a major change, since the original functionality remains unchanged and users will hardly notice it. The only thing it does is fix a bug related to window positioning, so this is a **fix** rather than a **feature**.

> * Will a user still be able to change data directly?

Users should not modify App's data directly; instead, they can modify the Svc's configuration file. However, users can forcefully modify the registry to change this data, though doing so serves no purpose.

> * Remember this needs to be documented.

Sure, but the code isn't actually that different.

Here's the current version: `localSettingsService.SetValue(“NotFirstRun”, true);` 
Here's the original version: `await localSettingsService.SaveSettingAsync(“NotFirstRun”, true);`

Jay-o-Way:
Can you tell/show me where the new data is being saved, and in what format?

ChenYiLins:
> Can you tell/show me where the new data is being saved, and in what format?

You can find more information [here](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/applicationdata/ApplicationData.md#35-unpackaged-app-data-stores); the data is saved in a special key-value format.

Jay-o-Way:
> > Can you tell/show me where the new data is being saved, and in what format?
> 
> You can find more information [here](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/applicationdata/ApplicationData.md#35-unpackaged-app-data-stores); the data is saved in a special key-value format.

Thanks. However, I only see the **location** in this paragraph. Would you like to copy and paste (a piece of) the raw data that is stored? Or simply attach the file?

ChenYiLins:
> Thanks. However, I only see the **location** in this paragraph. Would you like to copy and paste (a piece of) the raw data that is stored? Or simply attach the file?

These data are meaningless when viewed with the naked eye, so there's no need to worry about them:

image


image

Jay-o-Way:
> These data are meaningless when viewed with the naked eye, so there's no need to worry about them

That's exactly what I was pushing to. This means a user cannot simply edit the data in a text editor - in contrast to the JSON. This should be documented.

ChenYiLins:
> That's exactly what I was pushing to. This means a user cannot simply edit the data in a text editor - in contrast to the JSON. This should be documented.

I think you've misunderstood. We didn't start using JSON with the intention of letting users edit it; users shouldn't be able to access any App-related settings in the first place.

Jay-o-Way:
Ref: #839

ChenYiLins:
> Ref: #839

Are you referring to the backup and restore features? I don't think that's related to this PR; this should be a new PR.

Jay-o-Way:
I haven’t personally encountered the issues this PR addresses, so the need wasn’t immediately obvious to me. After reviewing the changes, it’s clear they focus on lifecycle stability and architectural cleanup, which makes sense within the WinUI migration. I tested the build and didn’t see any regressions in language switching, startup, or general behavior. From my perspective, the changes look safe and reasonable.

Tip: It might be worth double‑checking that all services and view models consistently use the centralized config instance to avoid any lingering initialization inconsistencies.

ChenYiLins:
> I haven’t personally encountered the issues this PR addresses, so the need wasn’t immediately obvious to me.

Yes, there was a previous issue where the window's size and position would occasionally not be saved, but I couldn't consistently reproduce the problem. It wasn't until Armin brought it up that I realized this wasn't just a minor issue happening on my computer.

Armin2208 (approved):
If it works it works is my statement

PR #1258, opened 2026-09-03
Use SettingsExpander.ItemsFooter for the InfoBar

Jay-o-Way:
Moved the InfoBar to the Footer, for simpler UI
image


ChenYiLins (approved):
LGTM

PR #1262, opened 2026-09-13
Feature: Switch to dark mode when enabling EnergySaver

ChenYiLins:
### Description

- Done one of #15

When the computer's power-saving mode is turned on, the theme switches to dark mode; when power-saving mode is turned off, it switches back to light mode.

@Armin2208 I didn’t place this feature in a module; instead, I put it in `SystemEventHandler` because I modeled it after the previous “Switch to dark mode when device is unplugged” feature. The original `RegisterThemeEvent` wasn’t prepared to accommodate any new features, so I made a few minor adjustments (without significantly altering the original structure), and I changed the registry listener to a subscription-based approach to align with the existing coding style.

Although it’s entirely possible to implement this feature as a module—I’ve tried it—doing so makes the structure look rather odd, since `DarkThemeOnBattery` is located in `Events` rather than in `Modules`.

### Supplement

The reason for not using the WinRT API is that, following the [Windows 11 24H2 update](https://learn.microsoft.com/en-us/windows-hardware/design/component-guidelines/energy-saver), the new power-saving mode features have been expanded, while `PowerManager.EnergySaverStatus` still uses the old battery power-saving mode. For more details, please refer to this [link](https://learn.microsoft.com/en-us/answers/questions/5579583/powermanager-energysaverstatus-not-working-properl). Testing has shown that when the device is plugged in, this value is always set to “Disabled.” The Win32 API behaves the same way. The registry solution, however, offers greater versatility, which is the compatibility that ADM has always sought.
### Screenshots

image


AutoDarkModeApp/Views/ConditionsPage.xaml
```diff
@@ -108,26 +108,32 @@
...
-                            Title="{helpers:ResourceString Name=AutoSwitchNotifyUnavailable_Title}"
                             IsClosable="False"
                             IsOpen="{x:Bind ViewModel.IsAutoSwitchNotifyUnavailableInfoBarOpen, Mode=OneWay}"
                             Message="{helpers:ResourceString Name=AutoSwitchNotifyUnavailable_Content}"
                             Severity="Informational" />
                     </controls:SettingsExpander.ItemsFooter>
                 </controls:SettingsExpander>
 
+                <!--  Energy Saver  -->
+                <controls:SettingsCard HeaderIcon="{ui:FontIcon Glyph=&#xF1E8;}">
```
Jay-o-Way: Hm, this icon does make sense because it's not about a physical battery. However, apparently Microsoft already tried to change/use this icon, but people for confused, so it actually reverted back to the battery icon. Consistency in the taskbar tray menu, etc... Now the icon is more like a "container" for energy settings.

Jay-o-Way, line 122: Hm, this 🍃 icon does make sense because it's not about a **physical** battery. However, apparently Microsoft already tried to change/use this icon, but people for confused, so it actually reverted back to the battery icon. Consistency in the taskbar tray menu, etc... Now the 🪫 icon is more like a "container" for energy settings.

This one seems best to me: image

ChenYiLins: Is ... that? It seems that on devices without a battery, this icon (P1) appears, while on devices with a battery, the battery icon (P2) appears. Alternatively, we could choose to enable different icons on different devices, though that would require adding some extra code. `Custom XAML Conditionals` might be helpful, but we’d need to upgrade the Windows App SDK to 2.0. I’ve already made the upgrade here: https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/pull/1253, but this PR is still awaiting approval.

P1:

image

P2:

image

Jay-o-Way: Hm, looks like it's more complicated. A few changes have been made (over time) for both desktop and laptop. Earlier comment only applies to devices with a battery.

AutoDarkModeApp/Views/ConditionsPage.xaml
```diff
@@ -108,26 +108,32 @@
...
                     Style="{StaticResource SettingsSectionHeaderTextBlockStyle}"
                     Text="{helpers:ResourceString Name=BatteryPowered}"
                     Visibility="{x:Bind ViewModel.BatterySettingsCardVisibility, Mode=OneWay}" />
-                <controls:SettingsCard
-                    ContentAlignment="Left"
-                    HeaderIcon="{ui:FontIcon Glyph=&#xE858;}"
-                    Visibility="{x:Bind ViewModel.BatterySettingsCardVisibility, Mode=OneWay}">
-                    <CheckBox Content="{helpers:ResourceString Name=BatteryDarkMode}" IsChecked="{x:Bind ViewModel.BatteryDarkModeEnabled, Mode=TwoWay}" />
+                <controls:SettingsCard HeaderIcon="{ui:FontIcon Glyph=&#xE858;}" Visibility="{x:Bind ViewModel.BatterySettingsCardVisibility, Mode=OneWay}">
+                    <controls:SettingsCard.Header>
```
Jay-o-Way: Why move this content into the header? Using content alignment is exactly as recommended in the Toolkit Gallery App.

AutoDarkModeApp/Views/ConditionsPage.xaml
Armin2208: The leaf icon is perfectly fine
Image

Armin2208:
>  I didn’t place this feature in a module; instead, I put it in SystemEventHandler because I modeled it after the previous “Switch to dark mode when device is unplugged” feature.

This is the right approach to do. Thank you very much, good choice!

> The reason for not using the WinRT API is that, following the [Windows 11 24H2 update](https://learn.microsoft.com/en-us/windows-hardware/design/component-guidelines/energy-saver), the new power-saving mode features have been expanded, while PowerManager.EnergySaverStatus still uses the old battery power-saving mode

Reading the linked forum post I need to say: that's a mess. How could be that the EnergySaver API returns wrong information and also different information depending on the battery percentage? WTF Microsoft?

It seems like there is an EnergySaver2 API 
https://learn.microsoft.com/en-us/uwp/api/windows.system.power.energysaverstatus2?view=winrt-28000

Why didn't you use that one? Because its not Windows 10 compatible? Is the registry key Windows 10 compatible? This is not a critique, just asking questions. 
But its important to know if the registry key is WIndows 10 compatible.

ChenYiLins:
> Why didn't you use that one? Because its not Windows 10 compatible? Is the registry key Windows 10 compatible? This is not a critique, just asking questions. But its important to know if the registry key is WIndows 10 compatible.

Unfortunately, I tested this on Windows 10, and that registry key doesn't even exist there—this solution is completely incompatible with Windows 10.

> It seems like there is an EnergySaver2 API https://learn.microsoft.com/en-us/uwp/api/windows.system.power.energysaverstatus2?view=winrt-28000

What a great discovery! This seems to be exactly what we need. However, after doing some research, I found that this API has only been registered but not officially enabled yet, and the pull request for this API was just pushed a week ago. See: https://github.com/microsoft/WindowsAppSDK/pull/6744. If we want to use this API, it means we’ll need to integrate the Windows App SDK into Svc. This API is indeed compatible with the new version of Windows 11. When combined with the existing `EnergySaverStatus`, it can ensure compatibility with Windows 10 and versions of Windows 11 prior to 24h2.

I am currently exploring solutions compatible with Windows 10 without having to introduce the Windows App SDK for the Svc.

Armin2208:
If the registry key solution is reliable, there is no need for the EnergySaver2 API. It's also weird that they introduce a second one.

There are two option for Windows 10 compatibility:
- fallback to energy saver API
- hiding this feature in UI

Why preserving windows 10 compatibility with ADM?
- still high market share 
- Still receiving updates 
- doesn't have native Auto dark mode, so a potential great user base.

ChenYiLins:
## Code quality: Compatible with Windows 10

### Description

Fall back to `EnergySaverStatus` to ensure compatibility with versions prior to 24H2.

However, please note that in previous versions, Power Mode had four settings, and energy-saving mode detection was triggered only when the device was set to the energy-saving setting.

### Screenshots

image

ChenYiLins:
## Code quality: Change the icon to ensure compatibility with Windows 10

### Description

@Jay-o-Way We have to revert to the battery icon because the leaf icon isn't supported in Windows 10.

Armin2208 (approved):

AutoDarkModeSvc/Monitors/AdmConfigMonitor.cs
```diff
@@ -204,14 +204,7 @@ public void PerformConfigUpdate(AdmConfig oldConfig, bool internalUpdate = false
...
     {
-        if (builder.Config.Events.DarkThemeOnBattery)
-        {
-            SystemEventHandler.RegisterThemeEvent();
-        }
-        else
-        {
-            SystemEventHandler.DeregisterThemeEvent();
-        }
+        SystemEventHandler.RegisterThemeEvent();
```
Spiritreader: `RegisterThemeEvent` isn't really fitting anymore as this function now both register and deregisters event hooks. Maybe we could name it sth like `UpdateThemeEventRegistrations` so it's clearer what it does.

Spiritreader (changes requested):
Everything looks good, just need to change the function name for legibility, then we're good to go!

Thanks @ChenYiLins!

Jay-o-Way:
Nit: `EBBB` is called "MobBatterySaver5". Even though visually that not a big deal, the name implies a very different intent. I think `EA95` is better. Otherwise: https://learn.microsoft.com/windows/apps/design/iconography/segoe-ui-symbol-font

AutoDarkModeSvc/Monitors/AdmConfigMonitor.cs
ChenYiLins: Done.

ChenYiLins:
> Nit: `EBBB` is called "MobBatterySaver5". Even though visually that not a big deal, the name implies a very different intent. I think `EA95` is better. Otherwise: https://learn.microsoft.com/windows/apps/design/iconography/segoe-ui-symbol-font

Done!

Spiritreader (approved):

